PR #643, opened 2026-04-11
fix: throw when selected model provider key is missing

### LINKED ISSUE #640: applyFallback silently falls back instead of throwing on missing API key
## Summary

Commit `adc7a8b7` (April 7, 2026) deliberately changed `applyFallback()` in `src/lib/model-fallback.ts` to **throw** when a provider's API key is missing, with the explicit rationale:

> "Never silently fall back to Anthropic (expensive) when a non-Anthropic model is configured but its API key is missing."

PAN-442's agent reverted this to a `console.warn` + silent fallback in commit `cdd56239` on the same day, as part of the Electron desktop app work — completely out of scope for that issue. The commit message was:

> "fix: restore fallback behavior in applyFallback when provider disabled"

Two subsequent PRs (PAN-462 in `5bc9a8f0`, PAN-510 in `e99e6546`) touched `applyFallback()` but perpetuated the silent fallback behavior, so it's still live on main today.

## Impact

When a user configures an OpenRouter/Google/Kimi/etc. model but hasn't set the API key, Panopticon silently charges their Anthropic key instead of telling them the key is missing. This is the exact scenario `adc7a8b7` was designed to prevent.

## Required Fix

Restore the throw behavior from `adc7a8b7`. The function should throw (or return a clear error) when a non-Anthropic model is requested but its provider key is missing — not silently swap to Claude.

Also verify the two companion fixes from the same commit:
- `agents.ts`: `getProviderEnvForModel` should throw for missing provider keys
- `cloister/specialists.ts`: same

## Files

- `src/lib/model-fallback.ts` — `applyFallback()` function (~line 156)
- `src/lib/agents.ts` — `getProviderEnvForModel`
- `src/lib/cloister/specialists.ts` — `getProviderEnvForModel`

## Key Commits

- `adc7a8b7` — original intentional fix (throw on missing key)
- `cdd56239` — PAN-442 agent reverted it
- `5bc9a8f0` (PAN-462) — reformatted, kept silent fallback
- `e99e6546` (PAN-510) — reformatted again, still silent fallback

## Origin

Collateral damage from PAN-442 (Electron desktop app). See PAN-639 for the related beads wipe from the same agent.

saschabuehrle:
## Summary
Restore fail-fast behavior for missing provider keys so Panopticon does not silently switch to Anthropic when a non-Anthropic model is selected.

## Changes
- `applyFallback()` now throws a clear error when the selected provider is disabled/missing API key
- updated `model-fallback` tests to assert throw behavior instead of fallback behavior
- removed warning-based silent fallback assertions

## Why
Silent fallback can cause unexpected provider usage and cost. This change makes configuration problems explicit and aligns model selection with user intent.

## Testing
- Updated `tests/lib/model-fallback.test.ts` expectations for missing-provider scenarios
- Test runner not executed locally in this environment (`vitest` not installed)

Fixes #640

Greetings, saschabuehrle

### DIFF
--- a/src/lib/model-fallback.ts
+++ b/src/lib/model-fallback.ts
@@ -2,8 +2,8 @@
  * Model Fallback Strategy
  *
  * When a non-Anthropic model is selected but its API key is missing,
- * automatically fallback to an equivalent Anthropic model. This ensures
- * Panopticon always works even without configuring external providers.
+ * fail fast with a clear error instead of silently falling back.
+ * This avoids unexpected Anthropic usage/cost when another provider is selected.
  */
 
 import { ModelId, AnthropicModel, OpenAIModel, GoogleModel } from './settings.js';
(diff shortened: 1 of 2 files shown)


eltmon:
Closing: stale with merge conflicts, will restart fresh if needed

PR #1001, opened 2026-05-07
fix: contracts editor.ts typecheck error (Schema.Literal → Schema.Literals)

eltmon:
Fixes TS2556 typecheck error in `packages/contracts/src/editor.ts`.

`Schema.Literal` in Effect v4 only accepts a single value. The existing code tried to spread an array into it, which fails under strict TypeScript. Switch to `Schema.Literals` which accepts an array, and replace the `.map() + as unknown` cast with an explicit typed tuple.

This fixes the typecheck failure discovered during PAN-951 verification.


eltmon (commented):
Review did not produce synthesis output

PR #1172, opened 2026-05-18
PAN-1170

### LINKED ISSUE #1170: TTS daemon lifecycle: install, start/stop, supervision
## Problem

PAN-829 wired the dashboard's in-process TTS subscriber (\`startTtsPlayback\`) but left the **Qwen3-TTS HTTP daemon** (\`skills/pan-tts/scripts/tts_daemon.py\`) as a manual-start, undocumented dependency. Every TTS feature in Panopticon — activity TTS playback, voice sampling, voice preset previews, AutoPreso voice feedback — POSTs to that daemon at \`daemonHost:daemonPort/speak\`. If the daemon isn't running, every TTS path fails with \`daemon-unavailable\` / \`TTS daemon unavailable\` and the user has no path-of-least-resistance to fix it.

Tonight a user hit:
- \"daemon is not running\" in the TTS status indicator
- \`{\"spoken\":false,\"result\":\"daemon-unavailable\",\"error\":\"TTS daemon unavailable\"}\` when trying to sample a voice preset

In both cases the actual fix was \`cd ~/Projects/qwen-tts && nohup .venv/bin/python /path/to/skills/pan-tts/scripts/tts_daemon.py &\`. That is not discoverable, and the dependency on a separate \`~/Projects/qwen-tts\` venv outside the panopticon-cli repo is even worse.

### 1. Install
- Move daemon Python deps under panopticon-cli (or under \`packages/qwen-tts-linux-x64/\` like the Moonshine sidecar) instead of \`~/Projects/qwen-tts/\`.
- Add \`pan install\` step that creates the daemon venv and installs \`qwen-tts\`, \`torch\` (with CUDA bindings), \`soundfile\`, \`numpy\`. Heavy — should be skippable via \`--skip-tts-daemon\` (heavy because of CUDA torch ~5GB).
- Optional: detect GPU and skip / warn if no CUDA available (daemon won't run usefully on CPU at the model's size).
- On non-Linux platforms, skip with a clear message.

### 2. Start / stop CLI
- \`pan tts start\` — launches the daemon (foreground or detached via \`--detach\`), waits for \`/health\` to return ok before exiting.
- \`pan tts stop\` — graceful SIGTERM → SIGKILL with PID tracked under \`~/.panopticon/pids/qwen-tts.pid\`.
- \`pan tts status\` — reports daemon PID, port, model, queue depth, uptime, GPU memory used.
- \`pan tts restart\` — wraps stop + start.
- All commands honor \`tts.daemonHost\` and \`tts.daemonPort\` from config.

### 3. Lifecycle integration
- \`pan up\` optionally launches the daemon (gated by \`tts.daemon.autoStart\` in config — default off because of the cold-start cost and VRAM footprint).
- Optional systemd user unit installer (\`pan tts install-systemd\`) for users who want auto-start on boot. Template lives in repo, \`pan install\` doesn't enable it by default.
- Dashboard supervisor / watchdog (extend the \`src/supervisor/watchdog.ts\` pattern from PAN-1139) detects daemon crash and restarts (with backoff and a give-up threshold so a misconfigured GPU doesn't restart-loop forever).

### 4. UI surface
- Dashboard Settings → TTS tab: \"Daemon: running on cuda:0 (4.2GB VRAM) | model loaded\" with a \"Start daemon\" button when it's down.
- Health badge in the sidebar (existing CloisterStatusBar already shows TTS — point it at the new status payload).

### 5. Docs
- Update \`skills/pan-tts/SKILL.md\` with the new CLI surface and remove the manual \`cd skills/pan-tts/scripts && python tts_daemon.py\` instruction (it survives as the lowest-level escape hatch).
- Add to \`docs/PANOPTICON_DEV_SOP.md\` (the doc PAN-1139 created) so post-restart troubleshooting includes \`pan tts status\`.

## Out of scope
- Replacing Qwen3-TTS with a different model.
- Cross-machine daemon hosting (the existing pan-tts SSE subscriber already handles that).
- Browser-side fallback synthesis (Web Speech API) for voice samples — keep the daemon as the single audio path so previews match what users hear in production.

## Acceptance

- [ ] Fresh \`pan install\` on a Linux/CUDA machine produces a working daemon (\`pan tts start\` → \`/api/tts/health\` returns ok).
- [ ] \`pan tts start\`, stop, status, restart all work and round-trip a single test phrase via \`pan tts test\`.
- [ ] Dashboard Settings TTS tab shows live daemon status and offers a one-click \"Start daemon\" when it's down.
- [ ] Crash + restart: kill the daemon process, watchdog brings it back within 30s.
- [ ] \`pan install --skip-tts-daemon\` completes without touching CUDA torch.
- [ ] Non-Linux platforms see a clear skip message during install (no failure).

## Why now

Tonight's session blocked at two separate points on the missing daemon — first \"TTS broken\" diagnosis, then \"voice sampling broken\" diagnosis — each requiring manual investigation. With \`pan tts start\` and the supervisor, both paths self-heal.

## Related
- PAN-829 (TTS lifecycle, merged) — shipped the in-process subscriber but didn't own the synth daemon.
- PAN-1139 (dashboard restart watchdog) — supervisor pattern to reuse here.
- PAN-1162 (ship-role allowHost) — example of a small substrate fix that unblocks a recurring failure mode.

eltmon:
Closes #1170

## Acceptance Criteria

- [ ] Implement issue

## Implementation Tasks

- [x] Implement issue


eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170
**Branch:** feature/pan-1170
**Review run:** agent-pan-1170-review-563f6d99
**Synthesized:** 2026-05-18

---

## Verdict: BLOCKED

5 distinct Tier-1 blockers found across 4 sub-reviews. All must be fixed before this PR can advance.

---

## Blocker 1 — Security: Unauthenticated TTS daemon HTTP endpoints (CRITICAL)

**Reviewer:** security
**Tier:** 1 / OWASP A01 Broken Access Control

The managed Qwen TTS daemon binds to `127.0.0.1:<port>` and exposes unauthenticated `POST /speak` and `POST /extract-embedding` with permissive CORS. Any attacker-controlled web page can POST to `http://127.0.0.1:8787/speak`, driving local audio output and triggering GPU model work without the user's consent.

**Root cause:** No credential or IPC protection on the daemon's HTTP interface.
**Fix required:** Add a daemon auth secret (generated under `PANOPTICON_HOME` with 0600 perms, passed via env), require it on every mutating endpoint, remove wildcard CORS, and gate browser access to the Panopticon dashboard origin only.

---

## Blocker 2 — Correctness: `/api/tts/health` lies about daemon state when TTS playback is disabled

**Reviewer:** correctness
**Tier:** 1

`checkTtsHealth()` returns `{ ok: false, running: false, error: 'tts disabled in settings' }` before ever calling `getTtsDaemonStatus()` when `runtimeConfig.enabled` is false. Fresh installs default to `tts.enabled: false`. This means `pan tts start` can start a healthy daemon but `GET /api/tts/health` still returns `ok: false`.

**Root cause:** The playback-enable toggle (`tts.enabled`) is being used as a daemon-liveness gate.
**Fix required:** Always query daemon liveness in `/api/tts/health`; surface `ttsEnabled` as a separate field instead of replacing daemon status.

*Also flagged identically in the requirements review (AC 1 partial failure).*

---

## Blocker 3 — Correctness: Watchdog ignores a manually-started daemon under default config

**Reviewer:** correctness
**Tier:** 1

`TtsWatchdog.tick()` returns early when `ttsConfig.enabled || ttsConfig.daemonAutoStart` are both false. `pan tts start` starts the daemon but does not set either flag. Default config has both false. So the watchdog never supervises a user-launched daemon.

**Root cause:** Watchdog activation is coupled to playback enablement rather than daemon presence.
**Fix required:** Decouple watchdog activation from playback settings — treat an existing pid/state file or an explicit supervision flag as activation signals.

---

## Blocker 4 — Performance: Concurrent daemon starts can spawn multiple processes

**Reviewer:** performance
**Tier:** 1

`startTtsDaemon()` checks `getTtsDaemonStatus()` then spawns without any process-wide lock. Overlapping calls from multiple dashboard tabs, CLI invocations, or the watchdog restart path can each see "not running" and each spawn a new Python daemon. Only the last pid is tracked.

**Root cause:** Check + spawn is not atomic; no lock serializes concurrent start requests.
**Fix required:** Add a pid-file or process-level lock under `PANOPTICON_HOME` shared by all callers (dashboard route, CLI, watchdog), so exactly one spawn occurs and subsequent callers observe the already-running state.

---

## Blocker 5 — Requirements: Watchdog timing does not guarantee 30s restart SLA

**Reviewer:** requirements
**Tier:** 1

Default watchdog polls every 10s, requires 2 consecutive failures before restart, then allows up to 30s for daemon health. In the worst case a killed daemon waits ~20s before restart is even attempted, plus up to 30s health wait = potentially > 50s.

**Root cause:** Recovery timing is not sized to the 30s acceptance criterion.
**Fix required:** Adjust poll interval, failure threshold, or health timeout so detection + recovery completes within 30s, or trigger an immediate first-failure restart within the acceptance window.

---

## Advisory (non-blocking)

**Performance:** Health polling forks `nvidia-smi` on every status check. Dashboard (10s), Settings (10s), and watchdog (10s) each independently invoke the driver tool. Split health (liveness from daemon `/health` endpoint + pid file) from GPU telemetry, and cache GPU memory sampling with a short TTL.

**Supply chain:** Python packages (`torch`, `qwen-tts`, `soundfile`, `numpy`) are installed unpinned during `pan install`. Prefer pinned versions with hashes or a committed requirements artifact.

---

## Summary

| Blocker | Reviewer | Description |
|---------|----------|-------------|
| 1 | security | Unauthenticated daemon HTTP endpoints (CSRF risk) |
| 2 | correctness + requirements | Health endpoint lies when TTS disabled |
| 3 | correctness | Watchdog ignores manually-started daemon |
| 4 | performance | Concurrent starts spawn multiple daemons |
| 5 | requirements | Watchdog timing exceeds 30s AC |

**All 5 blockers must be resolved before this feature can merge.**

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-563f6d99/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170 — 2026-05-18

## Verdict: **BLOCKED**

One correctness blocker must be resolved before this branch can advance.

---

## Per-Reviewer Summary

| Reviewer | Status | Blockers | Advisories |
|---|---|---|---|
| Security | PASSED | 0 | 0 |
| Correctness | BLOCKED | 1 | 1 |
| Performance | PASSED | 0 | 1 |
| Requirements | PASSED | 0 | 0 |

---

## Blocker — Watchdog restarts daemon after intentional `pan tts stop`

**File:** `src/supervisor/tts-watchdog.ts:93–120`
**Severity:** Correctness — breaks daemon lifecycle contract

**Root cause:** The watchdog uses `ttsConfig.enabled` as a "desired daemon running" signal. But `tts.enabled` is the TTS activity-speak toggle — it controls whether events are spoken aloud, not whether the daemon should be kept running. When a user runs `pan tts stop`:

1. `stopTtsDaemon()` removes the PID/state files (`tts-daemon.ts:442–471`).
2. On the next watchdog tick, `active` is still set to `true` because `ttsConfig.enabled` is unchanged.
3. Health check fails (daemon is down) → watchdog calls `startTtsDaemon()`.

Result: `pan tts stop` appears to succeed, then the supervisor brings the daemon back within the 5s poll interval. The daemon cannot be kept stopped while TTS is enabled in Settings.

**Fix required:** Separate crash recovery from intentional stop. The watchdog should restart based on `daemonAutoStart` or prior unexpected-crash state, not merely `tts.enabled`. A deliberate `stopTtsDaemon()` call should set a "manual stop" gate that the watchdog respects — e.g., a flag file or state field that `startTtsDaemon()` clears and `stopTtsDaemon()` sets, so the watchdog only auto-restarts when `daemonAutoStart` is true AND the manual-stop gate is not set.

---

## Advisory — Stale PID reuse

**File:** `src/lib/tts-daemon.ts:235`
**Severity:** Low (advisory, not blocking)

`isProcessAlive()` treats any PID that accepts signal 0 as the managed daemon. If `~/.panopticon/pids/qwen-tts.pid` survives a reboot or long downtime and the OS reuses that PID for an unrelated process owned by the same user, `startTtsDaemon()` will incorrectly believe the daemon is running, and `pan tts stop` can signal the wrong process. The fix should verify daemon identity (e.g., command line + start time) before trusting the PID, and clear stale state on mismatch.

---

## Advisory — `nvidia-smi` spawned on every status poll

**Files:** `src/lib/tts-daemon.ts:244,284`
**Severity:** Low (performance advisory)

`getGpuMemoryUsedMb()` runs `execFileAsync('nvidia-smi', ...)` on every status call. Polling consumers include: dashboard header (every 10s when TTS enabled), Settings page (every 10s), supervisor watchdog (every 5s), and CLI `pan tts status`. Multiple dashboard tabs multiply the work. This is acceptable at current scale but should be cached (TTL ~30s) or made opt-in before this becomes a higher-frequency dashboard/supervisor signal.

---

## Requirements Coverage

All 9 acceptance criteria are implemented. No gaps.

| AC | Status |
|---|---|
| Fresh `pan install` → working daemon | Implemented |
| `pan tts start/stop/status/restart` + `pan tts test` | Implemented |
| Dashboard Settings TTS tab: live status | Implemented |
| Dashboard Settings TTS tab: one-click Start daemon | Implemented |
| Crash + watchdog restart within 30s | Implemented |
| `pan install --skip-tts-daemon` skips CUDA torch | Implemented |
| Non-Linux platforms skip gracefully | Implemented |
| No blocking calls in dashboard server code | Implemented |
| JSONL / release tooling untouched | Not applicable |

---

## Security

No findings. Auth token generation, argument-safe subprocess calls, JSON type-checking, and origin validation are all correctly implemented.

---

## What Needs to Change

The watchdog lifecycle logic in `src/supervisor/tts-watchdog.ts` must be updated so that `pan tts stop` results in a daemon that stays stopped until either the user explicitly starts it again or a restart is warranted (crash + `daemonAutoStart` true). The "manual stop" gate concept described above is the recommended approach.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-0384327a/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170
**Branch:** feature/pan-1170
**Run ID:** agent-pan-1170-review-5273e5c6
**Synthesized:** 2026-05-18
**Verdict:** BLOCKED

---

## Verdict Summary

| Role | Status | Blockers | Notes |
|------|--------|----------|-------|
| Security | PASS | 0 | 2 non-blocking |
| Correctness | **BLOCKED** | 1 | 2 non-blocking |
| Performance | PASS | 0 | 1 non-blocking |
| Requirements | COMPLETE | 0 | 0 non-blocking |

**Overall: BLOCKED** — one correctness blocker must be fixed before this PR can advance.

---

## Gating Blocker

### `pan tts test` cannot speak on a fresh/default install
**Reviewer:** correctness
**File:** `src/cli/commands/tts.ts:183`
**Evidence tier:** Tier 1

`runTtsTest()` immediately reads `config.voice`, returns `no-voice` when it is empty, and only posts to the daemon after `findVoiceById()` resolves a saved voice. Fresh TTS config defaults `voice` to `""` (`src/lib/config-yaml.ts:677`), and the voice store returns an empty list when no `tts-voices.json` exists (`src/lib/tts-voices.ts:23`). The daemon can speak a direct preset (`Vivian` by default), but the CLI test path never falls back to that direct payload and no install path seeds a default saved voice.

**Impact:** On a fresh Linux/CUDA install, `pan tts start` can bring the daemon up, but `pan tts test` exits non-zero with `No system voice set — use pan tts voices set-default <name>` instead of round-tripping the test phrase. This breaks the primary smoke-test path for the new daemon lifecycle (acceptance criterion 2).

**Fix (advisory):** Make the smoke-test path usable without prior voice-library setup. Either seed a default preset voice and set `tts.voice` during install, or have `pan tts test` send a direct preset payload using the daemon's configured default preset when `config.voice` is empty.

---

## Non-Blocking Notes

### Security — Bound TTS request size and generation cost
**File:** `src/dashboard/server/routes/tts.ts:158`, `skills/pan-tts/scripts/tts_daemon.py:381`

`parseSpeakTtsInput()` accepts arbitrary-length `text`, `voice`, `instruct`, and arbitrary-length numeric `embedding` arrays before forwarding to the daemon. A same-origin dashboard context or local caller with the daemon token can submit very large bodies causing CPU/GPU memory pressure or long-running generation work. Fix: add explicit maximum JSON body size, maximum text/design/instruct lengths, and expected embedding shape/length validation at the dashboard boundary and again in the daemon.

### Security — Pin Python daemon dependencies installed by `pan install`
**File:** `src/lib/tts-daemon.ts:531`

`installTtsDaemonDependencies()` installs `torch`, `qwen-tts`, `soundfile`, and `numpy` without versions or hashes. Fresh installs resolve whatever versions the package indexes serve at install time, so behavior and vulnerability exposure can drift independently of the reviewed git revision. Fix: pin known-good versions via a checked-in requirements file with hashes or equivalent lock mechanism for the daemon venv.

### Correctness — Invalid TTS watchdog env values can create zero-delay poll/restart loop
**File:** `src/supervisor/tts-watchdog.ts:24`

`readTtsWatchdogConfig()` uses `Number(...)` directly for `pollMs`, `failThreshold`, `maxRestarts`, `windowMs`, and `startTimeoutMs`. Invalid env values become `NaN`, `0`, or negative numbers. In particular, `setTimeout(..., NaN)` behaves like a zero-delay timer, and `maxRestarts: NaN` makes the give-up guard ineffective. Fix: parse these env vars with the same positive-integer fallback pattern used by `src/supervisor/watchdog.ts:84`.

### Correctness — Stopping the watchdog during an in-flight tick can reschedule it
**File:** `src/supervisor/tts-watchdog.ts:80`

The timer callback always runs `void this.tick().finally(() => this.schedule())`; `stop()` only clears the current timer. If `stop()` is called while `tick()` is still awaiting status/start work, the `finally` callback schedules a new timer after shutdown. Fix: track a stopped/running flag and only call `schedule()` from the `finally` path when the watchdog is still started.

### Performance — GPU memory collection shells out on every health/status poll
**File:** `src/lib/tts-daemon.ts:298`

`getTtsDaemonStatus()` always awaits `getGpuMemoryUsedMb(pid)` inside `Promise.all`, and `getGpuMemoryUsedMb()` runs `execFileAsync('nvidia-smi', ...)` at `src/lib/tts-daemon.ts:258`. Health/status is coupled to an external GPU telemetry subprocess even when the caller only needs daemon liveness. The dashboard polls `/api/tts/health` every 10s and the supervisor watchdog polls status every 5s. Fix: split liveness from telemetry — e.g., `getTtsDaemonStatus(config, { includeGpuMemory: false })` for watchdog/health, and only include cached/refreshed GPU memory for UI detail or `pan tts status`.

---

## Clean Areas

- Dashboard TTS routes for health, start, speak, voice CRUD, and embedding proxying use async handlers and proper auth token inclusion.
- TTS daemon process state, auth-token creation (mode 0600), start lock, detached spawn, status, and stop paths are correct.
- Daemon calls use argument-array `spawn`/`execFile` rather than shell interpolation.
- Dashboard Settings TTS status/start UI wiring and daemon status polling are correct.
- Supervisor TTS watchdog restart flow, manual-stop gate handling, and the watchdog tick serialization guard are correct.
- Install path for `--skip-tts-daemon`, non-Linux/CUDA skips, and TTS dependency install branching are correct.
- All 7 acceptance criteria are implemented and covered by tests.
- No new synchronous blocking calls (`execSync`, `readFileSync`, etc.) in dashboard-server reachable paths.
- Public voice responses strip embeddings. Project-level TTS daemon host/port values are stripped before runtime merge.

---

## Recommendation

Fix the `pan tts test` blocker (use the daemon's default preset when `config.voice` is empty, or seed a default voice during install). The four non-blocking notes are hardening items that should be addressed but do not gate the PR.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-5273e5c6/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170

**Run:** agent-pan-1170-review-cccd170a
**Branch:** feature/pan-1170
**Date:** 2026-05-18

## Verdict: BLOCKED (2 blockers)

---

## Layer-by-Layer Verdicts

| Role | Verdict | Blockers | Advisories |
|---|---|---|---|
| Security | :white_check_mark: PASSED | 0 | 0 |
| Correctness | :x: BLOCKED | 1 | 1 |
| Performance | :x: BLOCKED | 1 | 1 |
| Requirements | :white_check_mark: PASSED | 0 | 0 |

---

## Security — PASSED

No security blockers or advisories. Local daemon auth, origin checks on dashboard TTS mutations, non-shell process spawning, and project-scope endpoint isolation all reviewed clean.

---

## Correctness — BLOCKED

### Blocker 1: Read-only status poll erases crash marker before watchdog can restart

**File:** `src/lib/tts-daemon.ts:311`
**Evidence tier:** Tier 1

`getTtsDaemonStatus()` calls `clearState()` whenever a PID file exists but `isProcessAlive(pid)` is false. The dashboard's `/api/tts/health` (polling every 10s via Settings + status bar) and the supervisor watchdog (polling every 5s) both call this. If a health poll wins the race against the watchdog after the daemon is killed, `clearState()` removes the PID/state files. The next watchdog tick sees `hasTtsDaemonState() === false`, marks itself inactive, and does not restart the daemon — leaving it down indefinitely instead of restarting within 30 seconds as required by AC4.

**Fix:** Make `getTtsDaemonStatus()` read-only — report stale PID/state without deleting it. Lifecycle owners (stop, foreground close, watchdog crash handling) should be the only callers permitted to clear state.

### Advisory 1: Async project config loader bypasses TTS endpoint stripping

**File:** `src/lib/config-yaml.ts:855`
**Evidence tier:** Tier 1

The sync loader strips `tts.daemonHost`/`tts.daemonPort` via `stripProjectTtsEndpoint()` before returning. The async loader (`loadProjectConfigAsync()`) returns the raw parsed YAML and does not apply the same strip, but both share `configCache`. If an async conversation-config route populates the cache first, project-scope TTS endpoint fields remain in the shared config, causing health/start to route to the wrong daemon depending on request order.

**Fix:** Apply `stripProjectTtsEndpoint()` in `loadProjectConfigAsync()` before returning, matching the sync path.

---

## Performance — BLOCKED

### Blocker 1: `/extract-embedding` bypasses bounded queue, creates unbounded GPU work

**File:** `skills/pan-tts/scripts/tts_daemon.py:434`
**Evidence tier:** Tier 1

`/extract-embedding` performs `load_design_model()`, `MODEL_DESIGN.generate_voice_design()`, `load_base_model()`, and `MODEL_BASE.create_voice_clone_prompt()` directly inside the `ThreadingHTTPServer` request thread — no `WORK_QUEUE.put_nowait()` or any backpressure gate. `/speak` correctly uses bounded queue + 429 on overflow; embedding extraction does not. At high concurrency, this accumulates blocked request threads holding sockets, exhausts daemon threads/memory, and makes `/health` and `/speak` contend with extraction traffic.

**Fix:** Gate embedding extraction behind the same bounded work queue as `/speak`, returning 429/503 when the queue is full.

### Advisory 1: `nvidia-smi` spawned for every health/status poll

**File:** `src/lib/tts-daemon.ts:258`
**Evidence tier:** Tier 1

`getTtsDaemonStatus()` unconditionally calls `getGpuMemoryUsedMb(pid)` which shells out to `nvidia-smi`. The dashboard status bar polls every 10s, Settings every 10s, and the watchdog every 5s — producing dozens of `nvidia-smi` invocations per minute at scale.

**Fix:** Cache GPU memory sampling with a short TTL, or split VRAM into a separate slow-path diagnostic endpoint.

---

## Requirements — PASSED

All 6 acceptance criteria are implemented and evidenced:

| # | Criterion | Status |
|---|---|---|
| 1 | Fresh install → working daemon | Implemented |
| 2 | start/stop/status/restart + test phrase | Implemented |
| 3 | Dashboard TTS tab live status + Start button | Implemented |
| 4 | Crash → watchdog restart within 30s | Implemented (watchdog 5s poll, 25s start timeout) |
| 5 | `--skip-tts-daemon` skips CUDA torch | Implemented |
| 6 | Non-Linux platforms see clear skip message | Implemented |

---

## Top Blockers (must fix before re-review)

1. **[Correctness] `getTtsDaemonStatus()` side-effect race** — AC4 crash-restart path fails when dashboard is open and `/api/tts/health` polls before the watchdog ticks. Make status reporting read-only.
2. **[Performance] `/extract-embedding` unbounded GPU work** — no backpressure, bypasses the bounded queue model that `/speak` correctly uses. Gate behind bounded queue with 429/503 on overflow.

---

## Passengers (advisories, non-blocking)

- Async config loader TTS endpoint stripping gap — request-order dependent misroute risk
- `nvidia-smi` per-poll overhead — additive process churn, not correctness-affecting

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-cccd170a/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170 — 2026-05-18

## Verdict: BLOCKED

The feature branch is correctly implemented against all 6 acceptance criteria and
passes security and correctness review. However, a **blocking performance finding**
must be addressed before the branch can advance to merge.

---

## Blocking Finding

### P1: Unbounded TTS request bodies and clone embeddings — CPU/GPU memory exhaustion

**Reviewer:** performance
**Evidence tier:** Tier 1
**File:** `skills/pan-tts/scripts/tts_daemon.py:428`

The Python daemon reads `Content-Length` directly and then `self.rfile.read(length)`
with no maximum before allocation. Queue backpressure is applied only *after* the
full request body has been read and parsed. A buggy or hostile local client can
send a huge JSON body or oversized clone embedding, forcing large JS/Python array
and GPU tensor allocation before any size limit rejects it.

At 10x load, or with a single oversized request, the daemon can exhaust process
memory or VRAM, taking down TTS playback and triggering cascading watchdog restarts.

**Required fixes:**
1. In the Python daemon: reject oversized `Content-Length` with HTTP 413 before `rfile.read`.
2. In the Python daemon: cap clone embeddings against expected Qwen speaker-embedding
   dimensionality with finite-number checks before storing or converting to a CUDA tensor.
3. In dashboard TTS routes: cap `request.text` payloads for TTS routes; cap `text`/`design`
   string lengths before sending to the daemon.

---

## Non-Blocking Advisories

| ID | Reviewer | Summary |
|---|---|---|
| S1 | security | Python TTS pip install path pulls unpinned dependencies at install time; ship a locked requirements file with exact versions and hashes |
| S2 | security | `/extract-embedding` failure branch returns raw ML exception text to authenticated callers; return a stable generic error code to clients, keep details server-side only |
| P1 | performance | Health polling can cause concurrent `nvidia-smi` invocations at cache expiry moments on multiple open Settings tabs; key an in-flight sampling promise by PID and have concurrent callers await it |

---

## Requirements Coverage: COMPLETE (6/6)

All acceptance criteria are implemented and verified by the requirements reviewer.

---

## Summary

| Dimension | Verdict |
|---|---|
| Requirements | PASS — 6/6 ACs implemented |
| Correctness | PASS — no issues |
| Security | PASS — 2 non-blocking advisories |
| Performance | **BLOCKED** — P1 unbounded allocation |
| **Overall** | **BLOCKED** |

The P1 allocation finding must be fixed and re-reviewed before the convoy can
advance to the test role.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-c261184e/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170
**Convoy run:** agent-pan-1170-review-a9b8a466
**Reviewers:** security ✓, correctness ✓, performance ✗ (1 blocker), requirements ✓
**Verdict:** BLOCKED
**Date:** 2026-05-18

---

## Blocker (1)

### Performance — TTS watchdog timer leak after stop
**File:** `src/supervisor/tts-watchdog.ts:80`
**Severity:** Blocker
**Evidence tier:** Tier 1

`stop()` only clears the currently scheduled timeout. The active timeout callback runs `void this.tick().finally(() => this.schedule())` unconditionally. If `stop()` is called while `tick()` is awaiting health checks, config reads, logging, or `startTtsDaemon()`, the `finally` block schedules a new timeout after `stop()` has cleared the old one. This leaves the watchdog active with a leaked polling timer even after callers requested shutdown.

**Impact:** The leaked timer keeps the supervisor event loop doing TTS health polling and restart attempts after shutdown/disable paths. At repeated supervisor lifecycle changes, this creates retained timers and unnecessary daemon status work every `pollMs`.

**Fix:** Track a `stopped`/`running` flag: set true in `start()` and false in `stop()`. Only call `schedule()` from the tick `finally` when the watchdog is still running. Also guard `schedule()` itself so it does not create a timer after stop.

---

## Non-blocking Advisories (1)

### Security — TTS route body cap before buffering chunked requests
**File:** `src/dashboard/server/routes/tts.ts:347`
**OWASP category:** A05: Security Misconfiguration

The route rejects large `Content-Length` values with `ttsPayloadTooLargeResponse()`, then calls `yield* request.text` and checks `bodyText.length` afterward. A non-browser client that can reach the dashboard can spoof a trusted `Origin` header and send a large chunked body without a truthful `Content-Length`, causing the server to buffer before the route-level size check runs.

**Fix:** If the Effect/Node HTTP layer does not already enforce a global request-body limit, read request bodies through a capped stream helper or configure a server-level max body size for these routes.

---

## Passing Reviews

| Role | Verdict | Blockers | Notes |
|------|---------|----------|-------|
| Security | PASSED | 0 | 1 advisory (above). Fixed command/argument arrays, daemon token uses 32 random bytes + 0o600, mutable dashboard routes apply origin checks, user inputs bounded. |
| Correctness | PASSED | 0 | Clean. All lifecycle paths (daemon, CLI, dashboard routes, Settings UI, config, watchdog) are consistent with changed call contracts. 43 targeted tests passed. |
| Requirements | PASSED | 0 | All 7 ACs implemented and verified. |

---

## Summary

PAN-1170 implements a full TTS daemon lifecycle with watchdog recovery, dashboard status UI, CLI commands, and cross-platform install guards. The functional scope is complete and correct. One performance defect must be fixed before merge: the watchdog can leak a polling timer if `stop()` races an in-flight `tick()`. The security advisory is a defense-in-depth improvement that should also be addressed.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-a9b8a466/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170 — agent-pan-1170-review-fb3799b9

## Verdict: **BLOCKED** (3 blockers across 2 roles)

---

## Blocker Summary

| # | Role | Severity | Title | Location |
|---|------|----------|-------|----------|
| 1 | Correctness | Blocker | Watchdog restart never replaces an unhealthy-but-alive daemon | `src/lib/tts-daemon.ts:369` |
| 2 | Performance | Blocker | Unvalidated watchdog timing env vars can create a tight supervisor loop | `src/supervisor/tts-watchdog.ts:24` |
| 3 | Correctness | Blocker | Healthy untracked daemons produce contradictory status and cannot be stopped | `src/lib/tts-daemon.ts:330` |

---

## Role-by-Role Summary

### Security — PASSED
No blockers. 2 non-blocking advisories:
1. **Unpinned Python packages** in TTS daemon installer — `torch`, `qwen-tts`, `soundfile`, `numpy` installed without pinned versions or hashes, creating supply-chain risk.
2. **Spoken text in daemon log** — utterance snippets are persisted to `~/.panopticon/logs/qwen-tts.log`, exposing TTS input beyond the playback moment.

Clean areas: origin checks, body bounds, token auth on daemon endpoints, no shell interpolation in spawn/CLI, no XSS in frontend TTS rendering.

### Correctness — BLOCKED
1 blocker, 1 advisory.

**BLOCKER C1 — Watchdog restart never replaces an unhealthy-but-alive daemon** (`tts-daemon.ts:369`)

`startTtsDaemon()` treats any live PID as "already running" and refuses to replace it. `existing.running` only means the PID file points at a live process — not that the daemon is healthy. When `/health` is failing but the Python process is still alive, the watchdog's restart action calls `startTtsDaemon()`, which sees `existing.running === true` and does nothing. The daemon remains wedged indefinitely.

Fix: either (a) have `startTtsDaemon()` short-circuit only when `existing.ok === true`, killing and replacing when `existing.running && !existing.ok`, or (b) expose a dedicated restart primitive the watchdog calls instead of `startTtsDaemon()`.

**ADVISORY C2 — Healthy untracked daemons produce contradictory status** (`tts-daemon.ts:330`)

If `/health` responds but pid/state files are missing or stale, status can be `{ ok: true, running: false, pid: null }`. The dashboard shows the daemon as running, but `pan tts stop` cannot stop it (it only reads the pid file). `pan tts start` may also spawn a duplicate because it only checks `existing.running`, not `existing.ok`.

### Performance — BLOCKED
1 blocker, 1 advisory.

**BLOCKER P1 — Unvalidated watchdog timing env vars create tight supervisor loop** (`tts-watchdog.ts:24`)

`readTtsWatchdogConfig()` passes raw `Number(...)` values directly to `setTimeout()`. `setTimeout(NaN)` or non-positive delays schedule immediately, turning the watchdog into a tight loop. `PANOPTICON_TTS_WATCHDOG_MAX_RESTARTS=abc` produces `NaN`, making `length >= NaN` always false — disabling the restart cap entirely. A single bad env value can make the supervisor continuously poll/restart instead of sleeping 5s between cycles.

Fix: reuse the existing positive integer parser from `src/supervisor/watchdog.ts` (or create a shared helper) and apply it with sane fallbacks and minimums before constructing `TtsWatchdogConfig`.

**ADVISORY P2 — Health polling waits on GPU telemetry child process** (`tts-daemon.ts:320`)

`getTtsDaemonStatus()` awaits `Promise.all([..., getGpuMemoryUsedMb(pid)])` on every health check. `getGpuMemoryUsedMb()` runs `nvidia-smi` with a 2s timeout when its 30s cache is cold, adding up to 2s latency to every cold health poll. The daemon's own `/health` is already available — GPU telemetry blocks the response unnecessarily.

### Requirements — COMPLETE
All 7 acceptance criteria are implemented and verified:
- AC1: Fresh install produces working daemon
- AC2: CLI start/stop/status/restart/test round-trip
- AC3: Dashboard TTS tab shows live status + one-click Start
- AC4: Watchdog restarts dead daemon within 30s
- AC5: `--skip-tts-daemon` avoids CUDA torch
- AC6: Non-Linux platforms see clear skip message
- AC7: No blocking sync calls in dashboard TTS request path

---

## Shared Root Cause

Blocks C1 and C2 share a theme: the daemon state machine conflates "PID exists" with "daemon is healthy/managed." C1 is the more critical gap because it breaks the crash-recovery AC (AC4). C2 is a secondary lifecycle gap.

Block P1 is independent but also in the supervisor path — it affects the same watchdog that services AC4.

---

## Fix Priority

1. **P1 (performance/watchdog env):** Fix env var validation first — a bad env value can make the supervisor spin continuously, affecting all TTS users.
2. **C1 (correctness/watchdog restart):** Fix the restart primitive to kill/replace unhealthy-but-alive daemons — this is the core crash-recovery AC.
3. **C2 (correctness/untracked daemon):** Reconcile daemon PID via `/health` response or expose unmanaged state explicitly.

Advisories S1, S2, P2 can be addressed in follow-up issues.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-fb3799b9/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170
**Branch:** feature/pan-1170
**Run:** agent-pan-1170-review-9c852f3d
**Synthesized:** 2026-05-18

## Verdict: BLOCKED

One blocking correctness issue prevents merge. Security and Performance both pass cleanly. All six acceptance criteria are implemented.

---

## Blocking Issue

### Watchdog restart loop: daemon model-loading is indistinguishable from a crash

**File:** `src/lib/tts-daemon.ts:401` and `src/supervisor/tts-watchdog.ts:113-133`

**Root cause:** `startTtsDaemon()` unconditionally SIGTERM's a live managed PID whenever `/health` is not OK, before spawning a replacement. The watchdog follows the same health path on every failed poll. The Python daemon only binds its HTTP server *after* `load_model()` completes (`tts_daemon.py:601-604`). On a fresh install or slow GPU, model load or first-time model download can take longer than the watchdog's 25s `startTimeoutMs` and the CLI/dashboard's 120s start timeout. Until `/health` responds, Panopticon classifies the process as crashed.

**Runtime consequence:** A live daemon that is still downloading/initializing the Qwen model is repeatedly SIGTERM'd before it ever serves `/health`. This creates a restart loop rather than bringing TTS online.

**Required fix (no workarounds):** Make the startup/loading state explicit. Options:
1. Have the daemon bind its HTTP server *before* `load_model()` and have `/health` return a distinct `{"status":"loading"}` JSON with the PID while the model is still initializing.
2. Store a `starting`/`initializing` state with a startup grace deadline in the daemon state file, and teach `getTtsDaemonStatus()` and `TtsWatchdog` not to replace a live managed PID until the startup window has expired.

Either approach requires a distinct "still initializing" state visible to the health/readiness layer before the model is loaded.

---

## Non-Blocking Notes

### Security: `/health` loopback-only exposure (low risk)
`tts_daemon.py:466` — `do_GET /health` is unauthenticated while POST routes require `X-Panopticon-TTS-Token`. With the default `daemonHost: '127.0.0.1'` this is local-only. If a future global config binds the daemon to a non-loopback interface, an unauthenticated network client could query model name, queue depth, and PID. No change required for current default; note for future non-loopback support.

### Correctness: Stale PID file can target a reused unrelated process (low risk)
`tts-daemon.ts:253` — `isProcessAlive()` treats `process.kill(pid, 0)` as proof the stored PID belongs to the TTS daemon. If the daemon exits and the OS reuses the PID before the next check, Panopticon can SIGTERM an unrelated process. Before terminating stale state, validate process identity (e.g., `/proc/<pid>/cmdline` contains the expected script path and start time matches state start time) or require a matching daemon health PID.

---

## Coverage Summary

| Criterion | Status | Notes |
|---|---|---|
| Fresh install → working daemon | Implemented | AC1 met |
| `pan tts start/stop/status/restart/test` | Implemented | AC2 met |
| Dashboard TTS tab with live status + one-click start | Implemented | AC3 met |
| Crash + watchdog restart within 30s | Implemented | AC4 met |
| `--skip-tts-daemon` skips CUDA torch | Implemented | AC5 met |
| Non-Linux skip message (no failure) | Implemented | AC6 met |

All 6 ACs are coded. The startup-distinguished-from-crash invariant is what prevents the feature from safely reaching users.

---

## Reviewer Scores

| Reviewer | Verdict |
|---|---|
| Security | PASSED |
| Correctness | BLOCKED |
| Performance | PASSED |
| Requirements | COMPLETE |

**Signal:** `pan admin specialists done review PAN-1170 --status blocked --notes "Watchdog restart loop: daemon model-loading is indistinguishable from a crash — fix requires explicit startup/loading state before /health is served"`

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-9c852f3d/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170 — 2026-05-18

## Verdict: BLOCKED (2 blockers)

---

## Blocker 1 — Correctness

**`pan tts start` reports failure while the daemon is still in its supported startup window**

`src/lib/tts-daemon.ts:428`

`DEFAULT_TTS_DAEMON_STARTUP_GRACE_MS` is 30 minutes and `getTtsDaemonStatus()` correctly returns `initializing` for a live managed daemon inside that window. However, `waitForTtsDaemonHealth()` loops only on `!latest.ok` and `startTtsDaemon()` returns `ok: status.ok` after the 120s wait timeout — the code explicitly models long model startup as valid state, but the public start contract converts that valid `starting` state into a failed result. A fresh install hitting `Qwen3TTSModel.from_pretrained()` (potentially multi-minute download/load for the 1.7B model) is the most likely trigger. The consequence is a false-failure exit from `pan tts start` and a misleading "Start daemon" offer in Settings during a legitimate startup.

**Fix:** Preserve `starting` as a successful/accepted lifecycle result. After the health wait times out, return `ok: true` with `status.phase === 'starting'` rather than `ok: false`, or return a distinct `starting`/`accepted` result. For callers that need to block to health, expose a separate blocking-start variant with a configurable deadline aligned to `startupDeadlineAt`.

---

## Blocker 2 — Performance

**TTS watchdog performs synchronous config I/O on every poll**

`src/supervisor/tts-watchdog.ts:104`

`tick()` loads TTS config synchronously with `loadConfig().config.tts` before each health check/restart decision. `loadConfig()` in `src/lib/config-yaml.ts` calls `existsSync`/`statSync`/`readFileSync` on every tick. This puts blocking filesystem and YAML parsing work into the supervisor's periodic event-loop path (default 5s cadence), sharing the event loop with `/health`, `/status`, `/tts-status`, and `/restart-dashboard`. At scale, on slow filesystems, or during config edits, this delays supervisor responses and watchdog scheduling.

**Fix:** Use an async config path (`loadConfigAsyncNoMigration()` or a new async full loader), keep mtime/config reads on `fs/promises`, and refresh the config outside the tick's critical path.

---

## Non-Blocking Advisories

| # | Area | Description |
|---|---|---|
| 1 | Security | Python TTS daemon inherits the full parent process environment — allowlist only required vars (`PATH`, `HOME`, locale/audio/CUDA vars, `QWEN_TTS_*` settings). (`src/lib/tts-daemon.ts:473, 531`) |
| 2 | Security | Runtime `pip install` in `installTtsDaemonDependencies()` uses unpinned packages — commit a pinned requirements file with hashes. (`src/lib/tts-daemon.ts:624`) |
| 3 | Correctness | Settings UI derives `unhealthy` from `ok ? online : running ? unhealthy : offline` and omits `phase`/`initializing` — renders a starting daemon as unhealthy and offers a redundant Start action. (`src/dashboard/frontend/src/components/Settings/SettingsPage.tsx:1045`) |

---

## Requirements Coverage: COMPLETE

All 6 acceptance criteria are implemented. No gaps.

| AC | Status |
|---|---|
| Fresh install → working daemon | Implemented |
| start/stop/status/restart + test round-trip | Implemented |
| Dashboard TTS tab live status + Start button | Implemented |
| Crash → watchdog restart within 30s | Implemented |
| `--skip-tts-daemon` bypasses CUDA torch | Implemented |
| Non-Linux shows clear skip message | Implemented |

---

## Summary for `pan admin specialists done`

```
2 blockers: (1) tts-daemon start returns false failure during legitimate 30-min startup grace window; (2) watchdog tick does sync config I/O blocking the supervisor event loop
3 advisories: env allowlist for daemon, pinned pip deps, Settings starting-state mapping
6/6 ACs implemented — requirements complete
```

---

*Synthesis by pan-review agent · Run agent-pan-1170-review-837a50e5*

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-837a50e5/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1170

# Review Synthesis — PAN-1170 — 2026-05-18

## Verdict: BLOCKED

Three blocking findings must be resolved before this PR can advance.

---

## Blocking Findings

### 1. PERFORMANCE — CUDA memory exhaustion from permanent multi-model retention

**File:** `skills/pan-tts/scripts/tts_daemon.py:141`
**Reviewer:** performance (Tier 1)

The TTS daemon preload policy causes three Qwen 1.7B bfloat16 models to accumulate permanently in GPU memory. `load_model()` (default), `load_design_model()` (lazy, called by `/extract-embedding` and design `/speak`), and `load_base_model()` (lazy, also called by those same paths) are stored as module globals with no eviction path. A single voice-design or clone request leaves all three resident for the remainder of the process lifetime.

**Impact:** On 8–12 GB GPUs (common developer machines), the first design/clone workflow can OOM the daemon or starve subsequent synthesis of VRAM, causing latency spikes, failed generation, or watchdog restart loops.

**Fix:** Implement a one-active-model policy — explicitly release the prior model (`del` + `gc.collect()` + `torch.cuda.empty_cache()`) before loading the next, or run design/embedding extraction in a short-lived worker subprocess.

---

### 2. REQUIREMENTS — AC1: `pan tts start` reports success before `/api/tts/health` is ok

**File:** `src/lib/tts-daemon.ts:536`, `src/lib/__tests__/tts-daemon.test.ts:99`
**Reviewer:** requirements (Tier 1)

`startTtsDaemon()` accepts `status.initializing === true` as a successful start outcome. `getTtsDaemonStatus()` returns `initializing` while the spawned process is alive but `/health` is not yet reachable during startup grace. The regression test explicitly locks this behavior in.

**Impact:** On a fresh Linux/CUDA install where Qwen model download or load takes longer than the start timeout, `pan tts start` exits 0 while `/api/tts/health` still returns `ok: false` / `phase: 'starting'`. The AC1 acceptance sequence (`pan tts start` → `/api/tts/health` returns ok) is not guaranteed.

**Fix:** Require `ok: true` for the default start success path, or explicitly surface an initializing/not-yet-healthy state to the caller and update the CLI/API contract accordingly.

---

### 3. CORRECTNESS — Stale PID state trusted before signaling

**File:** `src/lib/tts-daemon.ts:314`
**Reviewer:** correctness (Tier 1)

`isProcessAlive()` treats any `process.kill(pid, 0)` success as proof that the PID still belongs to the managed TTS daemon. `startTtsDaemon()` / `stopTtsDaemon()` then call `terminateProcess()` on that PID.

**Impact:** If a stale state file's PID has been reused by an unrelated process, `pan tts start`, watchdog restart, or `pan tts stop` can signal the wrong process. PID reuse is low-probability but high-consequence.

**Fix:** Record and validate process identity (e.g., `/proc/<pid>/cmdline` content or process start time) against state written at spawn before sending signals; clear stale state when identity does not match.

---

## Clean Reviews

| Reviewer | Status |
|---|---|
| Security | CLEAN — no findings |
| Correctness | CLEAN with 1 advisory (PID reuse above) |
| Performance | 1 blocking finding (CUDA model retention above) |
| Requirements | PARTIAL — 1 blocking finding (start health contract above), 5/6 fully implemented ACs |

---

## Summary

The TTS daemon feature is substantially implemented — all six acceptance criteria have evidence in the diff, and the security review found no issues. The three blockers are all in the Python daemon (`tts_daemon.py`) and the TypeScript start/stop lifecycle (`tts-daemon.ts`). None are cosmetic; all affect correctness under real-world conditions on the target Linux/CUDA deployment.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1170/.pan/review/agent-pan-1170-review-2c4bbd3c/synthesis.md

## Required action

`pan review request PAN-1170 -m "Fixed review issues"`

Issue #11, opened 2026-01-20
Add cost badge to Kanban cards


eltmon:
## Overview

Display a cost badge on each Kanban card showing the total AI cost for that issue.

**PRD Reference:** `docs/prds/reporting-prd.md` - Section 1.1
**Depends on:** #10 (Cost API endpoints)

## Background

Users need visibility into how much each issue costs in AI tokens. The Kanban board is the primary view, so costs should be visible at a glance.

### Visual Design

```
┌──────────────────────────────────────┐
│ ● MIN-123                      opus  │
│ Implement user authentication        │
│                                      │
│ 👤 Ed   🏷️ feature   💰 $2.47       │
│                                      │
│ [Watch] [Tell] [Kill]                │
└──────────────────────────────────────┘
```

### Cost Badge Styling

```tsx
// Color coding based on cost:
const getCostColor = (cost: number): string => {
  if (cost < 5) return 'text-green-400 bg-green-900/20';
  if (cost < 20) return 'text-yellow-400 bg-yellow-900/20';
  if (cost < 50) return 'text-orange-400 bg-orange-900/20';
  return 'text-red-400 bg-red-900/20';
};

// Badge component
<span 
  className={`inline-flex items-center gap-1 text-xs px-2 py-0.5 rounded ${getCostColor(cost)}`}
  title={`${sessions} sessions, ${formatTokens(tokens)} tokens`}
>
  💰 ${cost.toFixed(2)}
</span>
```

### Data Fetching

1. Fetch costs via new API endpoint: `GET /api/costs/by-issue`
2. Store in React Query cache with key `['costs', 'by-issue']`
3. Match costs to issues by `issueId`
4. Refresh every 60 seconds (costs don't change frequently)

### Implementation Steps

1. **Add cost data type** in `src/dashboard/frontend/src/types.ts`:
   ```typescript
   interface IssueCost {
     issueId: string;
     totalCost: number;
     tokenCount: number;
     sessionCount: number;
   }
   ```

2. **Add cost fetching** in `KanbanBoard.tsx`:
   ```typescript
   const { data: costs = [] } = useQuery({
     queryKey: ['costs', 'by-issue'],
     queryFn: async () => {
       const res = await fetch('/api/costs/by-issue');
       if (!res.ok) return [];
       const data = await res.json();
       return data.issues || [];
     },
     refetchInterval: 60000, // 1 minute
   });

   // Create lookup map
   const costMap = useMemo(() => {
[...]

Issue #10, opened 2026-01-20
Add cost API endpoints to dashboard server


eltmon:
## Overview

Add REST API endpoints to the dashboard server (`src/dashboard/server/index.ts`) to expose cost data to the frontend.

**PRD Reference:** `docs/prds/reporting-prd.md` - Section 4

## Background

The cost tracking system (`src/lib/cost.ts`) is complete but not exposed via the dashboard API. The frontend needs these endpoints to display costs on Kanban cards and issue details.

### Endpoint 1: GET /api/issues/:id/costs

Returns cost summary for a specific issue.

**Response Schema:**
```typescript
interface IssueCostResponse {
  issueId: string;
  totalCost: number;           // Total cost in USD
  totalTokens: number;         // Total tokens used
  currency: string;            // "USD"
  sessions: Array<{
    sessionId: string;
    startedAt: string;
    endedAt: string | null;
    type: string;              // "planning" | "implementation" | "review"
    model: string;             // "opus" | "sonnet" | "haiku"
    cost: number;
    tokens: number;
  }>;
  byModel: Record<string, number>;  // e.g., {"claude-opus-4": 0.85}
}
```

**Implementation:**
1. Import `readIssueCosts`, `summarizeCosts` from `../../lib/cost.js`
2. Call `readIssueCosts(issueId)` to get entries
3. Call `summarizeCosts(entries)` to aggregate
4. Transform to response schema
5. Return 200 with data or 404 if no costs found

### Endpoint 2: GET /api/costs/summary

Returns overall cost summary for today, week, and month.

**Response Schema:**
```typescript
interface CostSummaryResponse {
  today: CostSummary;
  week: CostSummary;
  month: CostSummary;
}

interface CostSummary {
  totalCost: number;
  entryCount: number;
  totalTokens: { input: number; output: number; total: number };
  byProvider: Record<string, number>;
  byModel: Record<string, number>;
}
```

**Implementation:**
1. Import `getDailySummary`, `getWeeklySummary`, `getMonthlySummary` from `../../lib/cost.js`
2. Call all three functions
3. Return combined response

### Endpoint 3: GET /api/costs/by-issue

Returns costs grouped by issue for dashboard display.

**Query params:**
- `days` (optional, default 30) - Number of days to look back

**Response Schema:**
```typescript
interface IssuesCostResponse {
  issues: Array<{
    issueId: string;
    totalCost: number;
    tokenCount: number;
    sessionCount: number;
  }>;
  period: {
[...]

PR #1225, opened 2026-05-19
PAN-1221

### LINKED ISSUE #1221: PAN-1148 follow-up: 2 minor review nits deferred at merge
Follow-up to PAN-1148 (#1184). Two minor review findings from cycle 16 (BLOCKED synthesis) were deferred at merge time because the work agent hit a kimi-k2.6 429 rate limit and the changes are low-risk:

### F1 — DELETE route emits wrong lifecycle event
**File:** `src/dashboard/server/routes/agents.ts:1069`

The shared `agentStopRoute` factory causes both `DELETE /api/agents/:id` and `POST /api/agents/:id/stop` to log `agent.stop_requested`. The original `deleteAgentRoute` logged `agent.delete_requested`. This is an observable behavioral regression in lifecycle logs — anything filtering on `agent.delete_requested` will miss DELETE events.

**Fix:** parameterize the lifecycle event name in the shared factory or split the routes back into distinct handlers.

### F2 — NaN-poisoning in reduce when startedAt is invalid
**File:** `src/dashboard/frontend/src/components/Agents/FleetAgentsView.tsx:290`

A `reduce` over agent durations doesn't guard against `startedAt` being undefined/invalid, which lets `NaN` poison the accumulator (final reduced value is `NaN` even though most inputs are valid).

**Fix:** add a `Number.isFinite()` check inside the reducer; skip invalid entries.

## Context

PAN-1148 was a 175/175 AC implementation across 113 files. Cycle 15 review passed. Cycle 16 surfaced these two new defensive-guard nits after the work agent pushed extra fixes for FleetAgentsView NaN handling. The fixes addressed most of cycle-15's perf advisories but introduced these specific guard gaps that cycle 16 caught.

Both fixes are tiny (single-line guards). Filed for clean follow-up since the main PR was already being held up by:
- A correctness reviewer recurring timeout pattern
- A kimi-k2.6 rate-limit halt mid-iteration
- The well-known test-passed-via-skip → no-ship-auto-dispatch substrate gap

The substrate gaps that caused this churn cycle are tracked separately (PAN-1215, plus a yet-to-be-filed issue for the test-skip-ship gap and the correctness-reviewer-keeps-crashing pattern).

---

## Update 2026-05-19 (post-merge UAT): Stream output panel shows "No recent stream output" for agents that have output

While doing light UAT on the merged PAN-1148 redesign, opened the IssueDrawer / sidebar popup for an active agent (e.g. PAN-1052 work agent) and the **stream excerpt panel renders "No recent stream output"** despite the agent actively emitting output (the agent had visible activity in the underlying tmux session and was making progress).

### Why this is suspicious

This is the same bug the cycle-cbe2c475 review identified pre-merge:

> **B1 correctness (run cbe2c475, blocker):** `DrawerActiveAgent.tsx:50` — The stream excerpt panel is built from `activityRail` (dashboard lifecycle events) instead of the agent's actual live output stream (`agentOutputById[activeAgent.id]`). Per PRD §4.7.8, this panel should show the agent's recent console output. Users see generic system messages instead of real agent output.

The work agent claimed to have fixed B1 in a later cycle, but observed behavior says otherwise. Either:
- The fix landed but reads from the wrong store key (the data is in `agentOutputById` but the component still selects via a stale selector / empty default).
- The fix landed but `agentOutputById` is not being populated server-side for these agents (subscription not active, snapshot not delivering output, etc.).
- The fix was reverted or the regression slipped past the cycle-15 audit.

### Fix should at minimum

- Reproduce: open Agents page → click any agent card whose underlying tmux session is producing output (e.g. PAN-1052 work agent) → observe drawer's stream excerpt panel.
- Confirm whether `selectAgentOutput(activeAgent.id)` returns data. If not, walk back: WebSocket subscription → read model → projection cache → which step is dropping it.
- Add a Playwright e2e test that asserts the stream excerpt panel renders the agent's last N lines (not the empty-state fallback) when `agentOutputById` has entries for that agent.

### Severity

User-visible regression on the merged redesign's headline feature. Worth fixing before the next ship.

eltmon:
Closes #1221

## Summary

PAN-1148 follow-up: three independent fixes deferred at merge or discovered in post-merge UAT.

### F1 — Restore `agent.delete_requested` lifecycle event for DELETE /api/agents/:id

**Problem:** PAN-1148 merged the DELETE and POST /stop routes into a shared `agentStopRoute` factory that hardcoded `agent.stop_requested` for both methods. DELETE was therefore emitting the wrong lifecycle event.

**Fix:** Parameterized the factory with a third `lifecycleEvent` argument so DELETE passes `'agent.delete_requested'` and POST /stop passes `'agent.stop_requested'`.

**Downstream consumer analysis (grep results):**
- `agent.delete_requested` — zero occurrences in `src/` or `packages/`. No consumer was subscribing to this event because it had never been emitted since the merge.
- `agent.stop_requested` — appears only at the single hardcoded site in `agents.ts` (now parameterized). No other code subscribes to or filters on this string.

**Conclusion:** No downstream consumers are broken by restoring the distinct event. No consumer updates are required.

### F2 — Guard FleetAgentsView avg-runtime reduce against NaN `startedAt`

**Problem:** When `startedAt` is `undefined`, `null`, empty, or malformed, `new Date(agent.startedAt).getTime()` returns `NaN`, which poisons the `avgRuntime` accumulator and produces `NaN` for the whole fleet.

**Fix:** Replaced the unguarded reduce with an IIFE that:
1. Maps `startedAt` to timestamps.
2. Filters to `Number.isFinite` values only.
3. Divides by the count of finite-duration agents (not `runningAgents.length`).
4. Returns `0` when no agent has a valid `startedAt` (matching the existing empty-array branch).

### F3 — Fix DrawerActiveAgent stream excerpt panel showing "No recent stream output" for active agents

**Root-cause diagnosis:**
The dashboard's data pipeline for agent output has four stages:
1. **Server emit** — something must capture agent output and append `agent.output_received` events to the event store.
2. **WS subscribe** — `EventRouter.tsx` subscribes to domain events over the WebSocket RPC stream.
3. **Store reduce** — `event-reducers.ts` handles `agent.output_received` by appending lines to `agentOutputById[agentId]`.
4. **Selector read** — `DrawerActiveAgent.tsx` calls `selectAgentOutput(activeAgent.id)` to read from the store.

Stages 2–4 were already implemented and working. Stage 1 was missing: **no server code was emitting `agent.output_received` events**. The reducer and selector had no data, so the stream-excerpt panel rendered the empty-state fallback.

**Fix (No Bandaids):** Added `AgentOutputService`, a background poller that captures tmux pane output for each running agent every 3 seconds and emits `agent.output_received` domain events when new lines appear. The fix is at the broken link (server emission) — not a defensive shim downstream in `DrawerActiveAgent.tsx`.

**Why not a downstream workaround?** Adding a fallback in the component (e.g., "if selector returns empty, try calling a REST endpoint") would mask the real gap, add latency, and create a second code path to maintain. Fixing the event pipeline at its source restores the architecture the dashboard was already designed around.

## Acceptance Criteria

- [x] Restore `agent.delete_requested` lifecycle event for DELETE /api/agents/:id
- [x] Guard FleetAgentsView avg-runtime reduce against NaN `startedAt`
- [x] Fix DrawerActiveAgent stream excerpt panel showing "No recent stream output" for active agents

## Tests

- F1: `src/dashboard/server/routes/__tests__/agents-lifecycle-event.test.ts` — 3 tests asserting DELETE vs POST /stop emit the correct lifecycle event.
- F2: `src/dashboard/frontend/src/components/Agents/__tests__/FleetAgentsView.test.tsx` — 3 regression cases (all valid, mixed valid/invalid, all invalid `startedAt`).
- F3: `src/dashboard/server/services/__tests__/agent-output-service.test.ts` — 11 tests covering `diffLines` algorithm and service integration.
- F3: `src/dashboard/frontend/tests/drawer-active-agent-stream.spec.ts` — Playwright e2e verifying the stream panel renders seeded output lines.


eltmon:
# Review CHANGES REQUESTED for PAN-1221

# Review Synthesis — PAN-1221 — 2026-05-20

## Verdict: **BLOCKED**

## Reviewer Results

| Sub-role | Status | Blockers |
|---|---|---|
| Security | PASS | 0 |
| Performance | PASS | 0 |
| Correctness | FAIL | 1 |
| Requirements | FAIL | 3 |

---

## Correctness — BLOCKER

### `splitLines` trailing newline breaks `diffLines` on every poll

**File:** `src/dashboard/server/services/agent-output-service.ts:32`

`text.split('\n')` produces a trailing empty string when the tmux pane output ends with a newline. This causes `diffLines` overlap detection to fail on every poll cycle where the agent has new output — instead of emitting only new lines, it emits the entire visible pane (up to 50 lines), flooding downstream consumers with duplicate data on every 3-second tick.

**Fix:**
```typescript
export function splitLines(text: string): string[] {
  const lines = text.split('\n')
  if (lines.length > 0 && lines[lines.length - 1] === '') {
    lines.pop()
  }
  return lines
}
```

This matches the pattern already used at `tmux.ts:762-763` for the same issue.

---

## Requirements — BLOCKERS (3)

All three are missing from the **PR description** (verified via `gh pr view`):

1. **AC #3 — Missing grep results and downstream consumer analysis**  
   The PR body must document `grep` results for `agent.delete_requested` and `agent.stop_requested`, confirming which consumers subscribe to each and that no downstream consumer is broken.

2. **AC #10 — Missing reproduction steps and root-cause diagnosis**  
   The PR body must explain the causal chain: no server code was emitting `agent.output_received` events → reducer and selector pipeline had no data → stream-excerpt panel showed empty state.

3. **AC #11 — Missing No Bandaids compliance statement**  
   The PR body must explicitly state that the fix adds `agent-output-service.ts` (the missing event emitter at the root cause) rather than adding defensive fallback code in `DrawerActiveAgent.tsx`.

---

## Clean Areas

- **Security:** No injection, auth bypass, XSS, or path traversal issues. `agentId` derives from filesystem enumeration, not HTTP input. WebSocket RPC filters by `agentId`. `capturePaneAsync` passes args as an array. SSH commands via `buildRemoteTmuxCommand` single-quote and escape each argument.
- **Performance:** All I/O is async. `withConcurrencyLimit(4)` bounds concurrent work. 50-line capture cap. `MAX_AGENT_OUTPUT_LINES` (200) bounds frontend state. Timer cleanup wired correctly. No regressions.
- **Route factory:** `createAgentStopHandler` correctly parameterizes lifecycle event. DELETE → `agent.delete_requested`, POST /stop → `agent.stop_requested`. No hardcoded strings.
- **FleetAgentsView NaN guard:** `Number.isFinite` filter and zero-denominator handling are correct.
- **Tests:** Vitest suite passes (383 files, 4243 tests). Lint clean. Typecheck clean.
- **Playwright spec:** Correctly structured; could not be executed (no running dashboard) — noted as non-blocking.

---

## Resolution Path

1. **Correctness blocker:** Apply the `splitLines` one-liner fix above.
2. **Requirements blockers:** Update the PR description with the three missing items (grep results, root-cause analysis, No Bandaids statement).
3. Re-run verification (typecheck, lint, test) after the code fix.
4. Re-review to confirm PR description is updated.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1221/.pan/review/agent-pan-1221-review-10126c27/synthesis.md

## Required action

`pan review request PAN-1221 -m "Fixed review issues"`

eltmon:
## Merge Blocked — Post-Rebase Verification Failed

Failed check: vbrief-ac

The branch was rebased successfully but verification failed. The work agent needs to fix the errors and resubmit.

eltmon:
## Merge Blocked — Post-Rebase Verification Failed

Failed check: vbrief-ac

The branch was rebased successfully but verification failed. The work agent needs to fix the errors and resubmit.

PR #1288, opened 2026-05-21
feat(deps): Effect migrate src/lib/remote/fly-api.ts (PAN-1249 slot-9)

eltmon:
## Summary

- Migrates `src/lib/remote/fly-api.ts` from Promise-based to Effect-native API
- `FlyApiError` converted from `extends Error` to `Data.TaggedError('FlyApiError')` with typed fields (`operation`, `statusCode`, `body`, `message`)
- All `FlyApiClient` methods now return `Effect.Effect<T, FlyApiError>` instead of `Promise<T>`
- `createFlyApiClient()` returns `Effect.Effect<FlyApiClient, ConfigError>` instead of throwing
- Test suite migrated to `@effect/vitest` with `it.effect()` throughout (17 tests, all green)
- Installs `@effect/vitest@4.0.0-beta.45` which was missing from node_modules

## Notes

`fly-provider.ts` callers have temporary typecheck errors — this is expected and will be fixed by the SCC migration task (`effect-migrate-cycle-activity-logger-and-13-more`) which explicitly includes `fly-provider.ts` in its scope and runs at a later wave.

## Test plan

- [x] All 17 tests in `tests/lib/remote/fly-api.test.ts` pass with `@effect/vitest`
- [x] `npm run typecheck` clean for `fly-api.ts` and test file
- [x] `npm run lint` clean

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -34,6 +34,7 @@
       "devDependencies": {
         "@commitlint/cli": "^20.5.0",
         "@commitlint/config-conventional": "^20.5.0",
+        "@effect/vitest": "4.0.0-beta.45",
         "@panctl/contracts": "workspace:*",
         "@types/better-sqlite3": "^7.6.13",
         "@types/inquirer": "^9.0.9",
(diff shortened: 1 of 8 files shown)


eltmon:
conflicts; re-dispatch

PR #1303, opened 2026-05-21
chore(pan-1249): runtimes/types.ts requires no Effect migration

eltmon:
## Summary

- `src/lib/runtimes/types.ts` is a pure TypeScript type/interface declarations file (zero function implementations, zero imports, zero I/O)
- The `void | Promise<void>`, `Agent | Promise<Agent>`, `boolean | Promise<boolean>` patterns in `AgentRuntime` interface method signatures are intentional polymorphic contracts that allow both sync and async implementors
- Changing the interface to `Effect.Effect<T, E>` here without simultaneously updating `claude-code.ts` and `pi.ts` (each in their own migration slots) would break TypeScript's structural subtyping check and fail `npm run typecheck`
- Same decision as `vbrief/types.ts` (commit `d811484e3`)

## Test plan

- [x] `npm run typecheck` passes (no changes needed)
- [x] `npm run lint` passes (no changes needed)
- No test file exists specifically for `src/lib/runtimes/types.ts`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -34,6 +34,7 @@
       "devDependencies": {
         "@commitlint/cli": "^20.5.0",
         "@commitlint/config-conventional": "^20.5.0",
+        "@effect/vitest": "catalog:",
         "@panctl/contracts": "workspace:*",
         "@types/better-sqlite3": "^7.6.13",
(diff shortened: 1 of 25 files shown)


eltmon:
conflicts

PR #1634, opened 2026-06-06
PAN-1395

### LINKED ISSUE #1395: Ctrl+K palette: semantic conversation search with excerpts (Phase 2)
## Goal

Phase 2 of the unified Ctrl+K palette: add **semantic search across Panopticon conversations** with **excerpt snippets that point to the matching message inside a JSONL session**. This is the heavy half of the palette redesign — Phase 1 (PR #TBD, "feat(palette): unified search") already wired memory FTS, observations, the `pan` command catalog, and issues into Ctrl+K.

This issue closes the loop on what the user asked for in the original prompt: "search pan commands, conversations (semantically.. where it will even show excerpt inside conversation that relates to it), issues, memories, observations, etc, ALL at once."

## Why Phase 1 didn't ship this

Conversation transcripts live as JSONL files under `~/.claude/projects/-home-eltmon-Projects/<project>/<sessionId>.jsonl` (and equivalents for other harnesses). Phase 1's memory FTS index doesn't cover them, and we deliberately don't ship a "keyword-only" conversation surface alongside memory results because it would create a half-built semantic story. Closed issues #729 and #185 describe the same intent and were never delivered — this issue picks up that thread with the additional Phase-1 surface as context.

### Backend
- **Conversation chunker**: stream a JSONL session, segment by message boundaries (and split long messages on ~512-token windows with overlap), emit `{ sessionId, projectId, role, ts, byteOffset, charLength, text }` records.
- **Embedding pipeline**: per-chunk embeddings, written to a sidecar SQLite (`~/.panopticon/conversations/embeddings.db`) with `sqlite-vec` (or comparable). Default model **OpenAI `text-embedding-3-small`** (matches the closed #729 spec). Pluggable provider via `~/.panopticon/config.yaml`.
- **Incremental indexing**: a watcher / on-demand reindex job that picks up new + appended JSONL files; idempotent on `byteOffset`.
- **Hybrid ranking**: BM25 (FTS5 over chunk text) **plus** vector cosine, fused by RRF — same shape as #185.
- **Search endpoint**: extend `GET /api/palette/search?q=&limit=` to add a `conversations` array of `{ sessionId, conversationId, projectId, role, ts, displayContent, excerpt, excerptSegments, rank }` whose `excerpt` uses the same `⦇…⦈` marker convention as memory excerpts so the frontend can render highlights without dangerously injecting HTML.
- **Open-conversation-at-message**: when a conversation result is selected, open the conversation drawer scrolled to the matching message (use `sessionId` + `byteOffset` to resolve).

### Frontend
- **CommandPalette**: add `Conversations` group above `Memory` in the result order. Reuse the existing excerpt-segment renderer.
- **Settings**: a new "Conversation Search" section exposing embedding provider/model/API-key, enable/disable toggle, last-indexed timestamp, and a "Reindex all conversations" button (cost-warning before kicking it off).
- **Cost guard**: estimate token cost before a full reindex and require explicit confirm if > $1.

### Tests
- Unit tests for chunker (boundary correctness, byte-offset accuracy after appended writes).
- Unit tests for the FTS+vector RRF ranker.
- Integration test for the search endpoint using a fixture session.

## Out of scope

- Replacing the `/` SearchModal flow for issues (that surface is fine; the palette continues to be the unified "everything" entry point).
- Indexing of non-conversation files (codebase semantic search is tracked separately).

## Acceptance criteria

- `Ctrl+K` → typing a phrase returns conversation hits **even when the original messages don't contain those exact words**, with excerpts pointing to the matched region.
- Selecting a conversation result opens its drawer scrolled to the matching message.
- Indexing happens incrementally; restarting the dashboard does not re-embed everything.
- Settings let me switch embedding provider/model and disable semantic search entirely.
- Cost of a full reindex is shown before running it.
- No regressions in the Phase 1 palette groups (commands / memory / observations / issues).

## Related

- Closed #729 — Feature: keyword + semantic search for conversations
- Closed #185 — Hybrid semantic search (BM25 + vector RRF)
- Phase 1 PR (this commit) — `feat(palette): unified Ctrl+K search` (commands + memory FTS + observations)

eltmon:
Closes #1395

## Acceptance Criteria

- [ ] Add conversationSearch config block (OFF by default)
- [ ] JSONL conversation chunker emitting byte-offset-keyed records
- [x] Embeddings sidecar DB: sqlite-vec + FTS5 schema and init
- [ ] Pluggable embedding provider (OpenAI default) + token-cost estimator
- [ ] Incremental indexer: scan, chunk, embed, upsert (idempotent)
- [ ] Live JSONL watcher feeding the incremental indexer
- [ ] Hybrid BM25 + vector ranker fused with RRF, with ⦇…⦈ excerpts
- [ ] Fill conversations[] in GET /api/palette/search
- [ ] Resolve sessionId + byteOffset to a message for open-at-message
- [ ] CommandPalette: Conversations group above Memory
- [ ] Selecting a conversation hit opens the drawer scrolled to the message
- [ ] Settings: Conversation Search section
- [ ] Reindex cost guard (>$1 requires explicit confirm)
- [ ] Unit tests: chunker boundaries + byte-offset across appends
- [ ] Unit tests: FTS + vector RRF ranker
- [ ] Integration test: palette search endpoint with a fixture session

## Implementation Tasks

- [x] Persist headless hook auto-approve scope
- [x] Fix remaining final test drift
- [x] Fix final quality gate test drift
- [x] Selecting a conversation hit opens the drawer scrolled to the message
- [x] Integration test: palette search endpoint with a fixture session
- [x] CommandPalette: Conversations group above Memory
- [x] Unit tests: FTS + vector RRF ranker
- [x] Fill conversations[] in GET /api/palette/search
- [x] Live JSONL watcher feeding the incremental indexer
- [x] Reindex cost guard (>$1 requires explicit confirm)
- [x] Hybrid BM25 + vector ranker fused with RRF, with ⦇…⦈ excerpts
- [x] Incremental indexer: scan, chunk, embed, upsert (idempotent)
- [x] Unit tests: chunker boundaries + byte-offset across appends
- [x] Settings: Conversation Search section
- [x] Pluggable embedding provider (OpenAI default) + token-cost estimator
- [x] Resolve sessionId + byteOffset to a message for open-at-message
- [x] Embeddings sidecar DB: sqlite-vec + FTS5 schema and init
- [x] JSONL conversation chunker emitting byte-offset-keyed records
- [x] Add conversationSearch config block (OFF by default)

<a href="https://capy.ai/project/8b56f615-9bcb-4db5-8b93-6447c9dd6bff/pull/1634"><picture><source media="(prefers-color-scheme: dark)" srcset="https://capy.ai/api/badge/open.svg?theme=dark"><source media="(prefers-color-scheme: light)" srcset="https://capy.ai/api/badge/open.svg">Open in Capy</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Semantic conversation search with excerpt previews and highlighted matches
  * Command Palette can surface conversation results and open a specific message in context
  * Conversation Search settings: enablement, provider/config, estimated reindex cost, and reindex action with confirmation

* **Chores**
  * Added embeddings sidecar DB and a filesystem watcher to keep conversation index in sync
  * Configuration schema and defaults for conversation search

* **Tests**
  * Extensive test coverage for indexing, embedding, ranking, watcher, API routes, and UI integration


eltmon:
# Review CHANGES REQUESTED for PAN-1395

# Review Synthesis — PAN-1395 — 2026-06-07T21:50:28Z

## Verdict: CHANGES REQUESTED

## Context
- Manifest: `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-0f80ae6c/context.json`
- Branch: `feature/pan-1395`
- Workspace: `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395`
- HEAD reviewed: `0f80ae6c616e2aacaddd614750ba369f7f6fb63b`
- Cycle number: 1
- Prior cycle SHA: none
- PR diff base: `aba8d713acdfc17fed84fbbe995d57c9a66a1f71`
- PR commits: 29
- Scope/convergence gates: all blocking findings below are in PR scope; cycle 1 means the cycle diff equals the PR diff.

## Convoy Status
| Sub-role | Signal | Output | Blocking findings |
| --- | --- | --- | --- |
| security | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-0f80ae6c/security.md` | 2 |
| correctness | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-0f80ae6c/correctness.md` | 3 |
| performance | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-0f80ae6c/performance.md` | 3 |
| requirements | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-0f80ae6c/requirements.md` | 3 |

## Blocking Findings

### [security] CSRFable reindex endpoint can upload local JSONL transcripts to the embedding provider — `src/dashboard/server/routes/settings.ts:717`
The new `POST /api/settings/conversation-search/reindex` route invokes `fullReindexConversationSearch()` without the dashboard mutation guard, trusted-origin/session/internal-token checks, JSON content-type enforcement, or CSRF validation. If conversation search is enabled, a cross-site form POST can force local Claude JSONL transcript chunks to be sent to the configured embedding provider and incur API spend. Gate this route with the existing dashboard mutation protections and move the >$1 confirmation to a server-validated nonce/confirmation flow.

### [security] Palette search exposes transcript excerpts through an unauthenticated route — `src/dashboard/server/routes/palette.ts:220`
The pre-existing `/api/palette/search` route now returns sensitive conversation-search excerpts and metadata from newly added PR code (`searchConversations()`/`toPaletteConversationHit()` and the `conversations` response group). Although the route itself predates the PR, the sensitive transcript disclosure is introduced by the changed conversation-search lines. Require dashboard authorization/origin validation before returning conversation hits, or omit the `conversations` group for unauthenticated requests.

### [correctness] Conversation palette hits use the Claude JSONL session id where the backend expects a Panopticon conversation name — `src/dashboard/server/routes/palette.ts:205`
`toPaletteConversationHit()` sets `conversationId` to the JSONL-derived Claude session id, and the frontend passes that value to `/api/conversations/:name/message-locator`. That route resolves only `getConversationByName(name)`, so normal conversations whose Panopticon name differs from their Claude session UUID fail with “Conversation not found” instead of opening the matched drawer message. Return a routeable conversation name in palette hits or make the locator route resolve by Claude session id.

### [correctness] Indexing advances the cursor past ignored trailing partial JSONL, permanently losing that message — `src/lib/conversation-search/indexer.ts:151`
`chunkConversationJsonl()` intentionally ignores trailing partial lines, but `indexConversationFile()` persists the cursor as `stat.size` even when those bytes were not safely consumed. If the writer later completes the JSONL line, the next incremental scan starts inside the now-complete line and skips it, causing silent permanent loss of that message from semantic search. Persist the last complete-line end offset instead of file size.

### [correctness] Enabled-to-enabled settings changes leave the watcher running with stale config — `src/dashboard/server/services/conversation-search-watcher.ts:112`
The settings PUT route saves a new conversation-search config and calls `syncConversationSearchWatcher()`, but `startConversationSearchWatcher()` returns the existing singleton whenever it is already active. The watcher captures DB/model/provider/API-key/roots in its constructor, so live indexing continues using the old configuration until a disable/restart cycle or dashboard restart. Restart or reconfigure the singleton when enabled settings change.

### [performance] Indexing buffers every new chunk from a JSONL file and embeds it in one unbounded request — `src/lib/conversation-search/indexer.ts:140`
`indexConversationFile()` collects all chunks from the new byte range into memory, sends one embeddings request for the entire array, and only then writes all rows and advances the cursor. Large transcripts can spike memory, exceed provider batch/token limits, and force repeated work after failures. Process bounded batches, persist each batch transactionally, and advance the cursor to the last complete consumed offset after each successful batch.

### [performance] Reindex cost estimation loads the entire transcript corpus into memory on a settings request — `src/lib/conversation-search/indexer.ts:104`
`estimateFullReindexConversationSearchCost()` pushes every chunk string from every discovered JSONL file into one `texts` array before estimating cost, and the settings route awaits this during an HTTP request. Cost estimation only needs aggregate counts, so this can allocate the whole searchable corpus at the moment users are sizing a full reindex. Stream chunks and accumulate token/cost totals without retaining all text.

### [performance] Ctrl+K conversation search performs synchronous SQLite open/schema/load/query work on the HTTP hot path — `src/dashboard/server/routes/palette.ts:220`
Every palette search opens the embeddings DB, performs synchronous filesystem/native setup and sqlite-vec loading, then runs synchronous `better-sqlite3` queries from the Node event loop. Because Ctrl+K search is debounced at interactive typing cadence, this can block unrelated HTTP/WebSocket/polling work and add visible palette latency. Move DB initialization/querying behind a long-lived service handle and off the request/event-loop hot path.

### [requirements] `byteOffset`/`charLength` do not accurately locate the source text bytes — AC5 / AC30 — `src/lib/conversation-search/chunker.ts:83`
The requirement asks for source-byte-accurate spans stable across appends, but `chunkConversationJsonl()` calculates `byteOffset` as JSONL line start plus decoded-text prefix length and `charLength` as JavaScript string length. It does not account for the byte position of the text payload within the JSON object, and the append test expects line-start offsets rather than verifying that slicing source bytes returns the chunk text. Track payload byte starts/byte lengths and update tests to assert source-byte slicing.

### [requirements] Full reindex mutates/scans while `conversationSearch.enabled` is false — AC13 — `src/lib/conversation-search/indexer.ts:74`
`fullReindexConversationSearch()` discovers transcript files, opens/creates the embeddings DB, and resets file cursors before delegating to the disabled-aware index path. This violates the disabled no-op acceptance criterion and can cause unnecessary re-embedding when the user later enables the feature. Add an early disabled return before discovery/DB/cursor mutation and cover it with a disabled full-reindex test.

## Non-blocking Findings

### [performance] Message-locator resolution reads large transcript files twice on selection — `src/dashboard/server/services/conversation-message-resolver.ts:32`
Selecting a hit first reads the full JSONL buffer to locate a sequence and then reparses the transcript via `parseConversationMessages()`. This is user-triggered rather than a high-frequency path, but large transcripts can still pay avoidable latency and memory pressure. Consider streaming to the target offset and reusing parse/cache behavior around the matching sequence.

### [performance] Settings status polling repeatedly opens the embeddings DB even when search is disabled — `src/dashboard/server/routes/settings.ts:684`
The Settings page polls status every 30 seconds, and the backend opens the embeddings DB unconditionally after reading config. When `conversationSearch.enabled` is false, return disabled status without DB open/schema/extension work.

## Clean Sub-roles
None.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-0f80ae6c/synthesis.md

## Required action

`pan review request PAN-1395 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1395

# Review Synthesis — PAN-1395 — 2026-06-07T23:04:00Z

## Verdict: CHANGES REQUESTED

## Context
- Manifest: `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/context.json`
- Branch: `feature/1395`
- Workspace: `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395`
- HEAD reviewed: `fe887e315a7fb1e6ab5fba54d964afccbb610fac`
- Cycle number: 2
- Prior cycle SHA: `0f80ae6c616e2aacaddd614750ba369f7f6fb63b`
- PR diff base: `f8da79bfda90fc88bea39d52b83e022177f77ac6`
- PR commits: 35
- Scope/convergence gates: all blocking findings below cite lines confirmed inside the PR diff. Cycle 2 does not trigger the cycle-diff convergence gate.

## Convoy Status
| Sub-role | Signal | Output | Blocking findings |
| --- | --- | --- | --- |
| security | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/security.md` | 0 |
| correctness | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/correctness.md` | 2 |
| performance | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/performance.md` | 1 |
| requirements | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/requirements.md` | 2 |

## Blocking Findings

### [correctness] Palette conversation hits can lose the target-message locator when Command Deck re-applies the deep link — `src/dashboard/frontend/src/App.tsx:550`
`handleOpenConversationHit` creates the target-bearing pane under the `projectId` from the indexed JSONL hit and then calls `setConversationRoute()`. Command Deck's `/conv/:id` effect can then resolve the same conversation to the registered project deck and open/focus a second pane without `targetMessageId`, `targetMessageIndex`, or `targetMessageNonce`. Selecting a Ctrl+K conversation result can therefore navigate to the conversation but fail to scroll/outline the matched message. Route palette hits through the same registered-project resolution path, or preserve the locator through the deep-link application.

### [correctness] Conversation search reports available when the embedding provider is unavailable — `src/dashboard/server/routes/settings.ts:726`
The status route reports availability from `db.available` only, while `createConversationEmbeddingProvider()` can return an unavailable provider when the configured API-key env var is missing. With `conversationSearch.enabled: true` and no usable API key, Settings can show `available: true`, the watcher can start, and indexing/searching then silently fail or return empty results. Provider availability and its reason must participate in status, indexing, and watcher-start gating.

### [performance] Conversation-search watcher can launch unbounded overlapping indexers — `src/dashboard/server/services/conversation-search-watcher.ts:105`
`schedule()` debounces file events only until the timer fires, then launches `void this.indexFile(...)` with no global concurrency limit, per-file in-flight guard, or serialized DB-write queue. Later appends to the same file or bursts across active conversations can create concurrent embedding batches and synchronous SQLite writes, causing duplicated embedding work, rate-limit spikes, `SQLITE_BUSY` contention, and avoidable cost. Add a bounded indexing queue and per-file coalescing/rerun behavior.

### [requirements] `byteOffset`/`charLength` do not accurately locate source bytes for escaped or multipart JSONL content — AC5 — `src/lib/conversation-search/chunker.ts:99`
The acceptance criteria require byte-accurate source locators stable across appends. The chunker finds a JSON string start and advances by decoded-text byte lengths, which diverges from raw JSONL byte offsets for escaped newlines, quotes, backslashes, and other escapes. Multipart content is concatenated from decoded text leaves but anchored at the first text leaf, so the stored span can cross non-contiguous source bytes. Track raw JSON string spans while decoding or emit chunks only from contiguous source leaves, and add regression tests for escaped and multipart content.

### [requirements] Conversation Search settings do not provide an editable persisted API key — AC25 — `src/dashboard/frontend/src/components/Settings/SettingsPage.tsx:2177`
The requirement says provider/model/API-key and enable toggle are editable and persist. The UI exposes an "API key env var" input and the API persists `apiKeyRef`, while the embedding provider reads only `process.env[apiKeyRef]`. A user cannot configure the API key through the persisted Settings surface, and existing persisted OpenAI API keys are not consumed by this provider path. Add/wire a persisted secret/API-key setting, or explicitly resolve `apiKeyRef` to supported config keys in addition to environment variables, and update UI/tests to match.

## Non-blocking Findings

### [security] ~ Conversation-search status/estimate endpoints expose transcript-corpus metadata without dashboard auth — `src/dashboard/server/routes/settings.ts:708`
The new status and reindex-estimate GET routes do not enforce the same dashboard auth/origin checks as palette search before returning DB path, indexing stats, and corpus scan/cost metadata. This does not directly expose message contents, so it remains advisory, but the endpoints should require dashboard auth before touching private transcript/indexing surfaces.

### [performance] ~ Message-locator resolves a hit by reading/parsing the whole transcript twice — `src/dashboard/server/services/conversation-message-resolver.ts:32`
Resolving one byte offset first loads the whole JSONL into a `Buffer`, then parses the transcript again to map the sequence. This is user-triggered rather than a hot indexing path, but large Claude transcripts can incur avoidable latency and memory pressure. Stream to the target offset and/or reuse parser/cache behavior.

### [performance] ~ Conversation palette result mapping does per-hit synchronous conversation lookups — `src/dashboard/server/routes/palette.ts:206`
`toPaletteConversationHit()` synchronously resolves the routeable conversation name for each ranked hit, so multiple chunks from one session can repeat the same lookup up to the request limit. Cache or bulk-resolve unique session IDs per request.

### [performance] ? Reindex cost estimate rescans and rechunks every JSONL file on each request — `src/lib/conversation-search/indexer.ts:108`
The settings estimate path performs a full corpus scan on demand. It is admin/settings-only and default-disabled, but large corpora can still make the action slow. Cache by file size/mtime or run estimation as a background job.

## Clean Sub-roles
None.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/synthesis.md

## Required action

`pan review request PAN-1395 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1395

# Review Synthesis — PAN-1395 — 2026-06-08T15:35:23Z

## Verdict: CHANGES REQUESTED

## Context
- Manifest: `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/context.json`
- Branch: `feature/pan-1395`
- Workspace: `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395`
- HEAD reviewed: `fe887e315a7fb1e6ab5fba54d964afccbb610fac`
- Cycle number: 2
- Prior cycle SHA: `0f80ae6c616e2aacaddd614750ba369f7f6fb63b`
- PR diff base: `f8da79bfda90fc88bea39d52b83e022177f77ac6`
- PR commits: 35
- Scope/convergence gates: cycle 2 does not trigger the cycle-diff convergence gate. The 6 blocking findings below cite newly added PR files/routes/functions. The security reviewer’s loopback-session finding was demoted because the cited trust condition existed at the PR diff base and the cited line is not a PR-introduced line.
- Proportionality: 6 synthesized blockers is below the 105-blocker threshold (3 × 35 commits), so no proportionality deferral was applied.

## Convoy Status
| Sub-role | Signal | Output | Blocking findings |
| --- | --- | --- | --- |
| security | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/security.md` | 0 |
| correctness | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/correctness.md` | 3 |
| performance | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/performance.md` | 3 |
| requirements | ready | `/home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/requirements.md` | 0 |

## Blocking Findings

### [correctness] Chunk offsets can point at JSON metadata instead of the message payload — `src/lib/conversation-search/chunker.ts:224`
`extractMessagePayloads()` finds each encoded text leaf with `rawLine.indexOf(encoded, searchFrom)`, searching from the beginning of the full JSONL line instead of from the specific `message.content` leaf. If message text appears earlier as JSON metadata or another field value (for example `user`, `assistant`, or `text`), the stored `byteOffset`/`charLength` can locate the wrong source bytes. That violates the byte-offset contract used by indexing, resolving, and debugging. Derive offsets from the specific content leaf/token span and add regression coverage for text values that collide with earlier metadata.

### [correctness] Tool-result chunks are indexed but cannot be opened from palette results — `src/lib/conversation-search/chunker.ts:239`
`extractTextLeaves()` indexes any object with a string `content` property, including Claude `tool_result` blocks. The message locator maps offsets to JSONL sequences and then searches rendered `parsed.messages`, but pure tool-result lines are rendered as work-log entries rather than `ChatMessage` rows. Selecting a semantic hit from tool output can therefore return 404 instead of opening/scolling the drawer. Align indexed chunks with resolvable UI targets: either skip non-message tool results or extend locator/timeline targeting to work-log rows.

### [correctness] Disabling the watcher does not stop in-flight startup indexing — `src/dashboard/server/services/conversation-search-watcher.ts:80`
`ConversationSearchWatcher.start()` launches `indexAll()` fire-and-forget, while `stop()` only clears queued timers and closes chokidar. If `conversationSearch.enabled` is turned off while a startup scan is already running, that scan continues embedding transcripts and writing the DB after the feature is disabled. Track/cancel/await active startup and incremental indexing work, and propagate an abort/stop signal through the indexer and embedding provider before reporting the watcher stopped.

### [performance] Long-message chunking rescans each payload from the beginning for every window — `src/lib/conversation-search/chunker.ts:100`
For every split window, `chunkConversationJsonl()` calls `encodedSpanForDecodedSlice()`, which starts at `rawIndex = 0` and walks the encoded payload again. Large transcript messages split into many windows therefore pay superlinear offset-resolution cost; the line reader also repeatedly `Buffer.concat`s while waiting for a newline. Precompute decoded-to-encoded offsets once per payload or carry a monotonic encoded cursor across ordered windows, and accumulate large line chunks without repeated full-buffer copies.

### [performance] Selecting a conversation hit loads the entire JSONL transcript into memory and then reparses it — `src/dashboard/server/services/conversation-message-resolver.ts:50`
`findJsonlSequenceForByteOffset()` reads the whole transcript into a `Buffer`, then `resolveConversationMessageLocator()` calls `parseMessages(sessionFile)` to parse the transcript again before the conversation pane loads messages. A single Ctrl+K result click now scales with full transcript size instead of the target offset, creating avoidable latency and memory pressure for large JSONL sessions. Resolve by streaming only until the requested byte offset, storing locators with chunks, or reusing existing parse/cache state.

### [performance] Full-corpus indexing has no singleflight guard and can duplicate expensive reindex jobs — `src/dashboard/server/routes/settings.ts:796`
The manual reindex route estimates and then runs `fullReindexConversationSearch()` directly, while the watcher can also be running a startup `indexAll()` scan. There is no shared in-progress flag or coordinator across manual reindex, startup scan, and watched-file indexing, so multiple tabs/retries/manual reindex during startup can launch duplicate corpus walks, embedding calls, and SQLite writes. Add a process-wide indexing coordinator/singleflight so manual reindex returns or rejects against the current job and startup/incremental indexing is serialized or queued appropriately.

## Non-blocking Findings

### [security] [demoted: pre-existing] Loopback requests can mint durable dashboard sessions without the internal token — `src/dashboard/server/routes/dashboard-auth.ts:184`
The security reviewer flagged the loopback trust condition in `rejectUnauthorizedDashboardSessionMintRequest()`. Scope confirmation showed that the `!hasDashboardInternalToken(request) && !isLoopbackPeer(request)` condition existed at the PR diff base (`f8da79bfda90fc88bea39d52b83e022177f77ac6`) and the cited line is not PR-introduced. The risk remains worth addressing in a dedicated issue, but it is out of PR scope for this review cycle.

### [security] ? Conversation-search config accepts privileged path/key references with little runtime validation — `src/lib/settings-api.ts:757`, `src/lib/config-yaml.ts:2118`, `src/lib/config-yaml.ts:2122`
Constrain `dbPath` to the Panopticon data directory and whitelist/validate `apiKeyRef` aliases to reduce blast radius if settings mutation is ever exposed too broadly. This is hardening advice, not a merge blocker.

### [correctness] ~ Search provider cache ignores API-key value changes — `src/dashboard/server/services/conversation-search-service.ts:14`
The conversation-search service signature includes `apiKeyRef` but not the resolved API key value. If the user rotates/fixes `api_keys.openai` while leaving `apiKeyRef` unchanged, the cached provider can keep using the old captured key until an unrelated service reset or restart. Include a resolved-key fingerprint/version in the signature or close the conversation-search service after settings saves that can change provider credentials.

### [performance] ~ Ctrl+K waits for memory FTS before starting semantic conversation search — `src/dashboard/server/routes/palette.ts:316`
Starting the conversation-search promise before awaiting memory-search fanout would avoid putting memory-search latency in front of the embedding/vector path. The current ordering is inefficient but not blocking.

### [performance] ~ Reindex POST recomputes the full estimate after the frontend already estimated it — `src/dashboard/server/routes/settings.ts:777`
The frontend estimates reindex cost and then the POST route performs a second full estimate before reindexing. Persist a short-lived signed estimate/nonce snapshot or background estimate job so the POST can validate confirmation without rescanning the corpus. This is secondary to the singleflight blocker.

## Clean Sub-roles
- requirements — reported all 35 in-PR-scope requirements implemented and no findings.

Source: /home/eltmon/Projects/panopticon-cli/workspaces/feature-pan-1395/.pan/review/agent-pan-1395-review-fe887e31/synthesis.md

## Required action

`pan review request PAN-1395 -m "Fixed review issues"`

PR #1940, opened 2026-06-16
fix(db): prevent schema migration from downgrading user_version (PAN-1932)

### LINKED ISSUE #1932: Schema migration downgrades user_version when DB is newer than code (=== SCHEMA_VERSION should be >=)
## Summary

The DB schema-migration entry point returns early only when `currentVersion === SCHEMA_VERSION`. If a database created by a **newer** build (higher `user_version`) is opened by older code, the early-return is skipped, no migration branch matches, and the function falls through to re-stamp `user_version` back down to the older `SCHEMA_VERSION` — silently **downgrading the schema-version metadata**.

This is **pre-existing** (predates PAN-1847 by many schema versions) and was surfaced by CodeRabbit on PR #1930 against the regenerated cost-event hook, which merely mirrors the source pattern.

## Root cause

`src/lib/database/schema.ts:744`:

```js
if (currentVersion === SCHEMA_VERSION) {
  return;
}
```

When `currentVersion > SCHEMA_VERSION` (older code, newer DB): this guard is false, every `if (currentVersion < N)` migration branch is false, and the trailing `db.pragma('user_version = ${SCHEMA_VERSION}')` (e.g. `schema.ts:1619` on the PAN-1847 branch) rewrites the version **downward**.

The same logic is embedded (generated) in `sync-sources/hooks/record-cost-event.js` (CodeRabbit pointed at the `=== 56` / re-stamp pair there) — fixing the source and regenerating the hook covers both.

## Fix

Make the early-return cover "at or ahead of what this code knows":

```js
// before
if (currentVersion === SCHEMA_VERSION) {
// after
if (currentVersion >= SCHEMA_VERSION) {
```

Then regenerate `sync-sources/hooks/record-cost-event.js` from source — **do not hand-edit the generated hook.**

(Find by code: search for `currentVersion === SCHEMA_VERSION` in `src/lib/database/schema.ts`.)

## Acceptance

1. Opening a DB whose `user_version` is greater than `SCHEMA_VERSION` leaves `user_version` **unchanged** (no downgrade) and runs no migration branch.
2. Opening a DB at exactly `SCHEMA_VERSION` still returns early (unchanged behavior).
3. Opening an older DB still migrates forward and stamps `SCHEMA_VERSION` (unchanged behavior).
4. The regenerated `record-cost-event.js` reflects the `>=` guard; no hand-edits.

Severity: low (only triggers on app-version downgrade against a newer DB), but it corrupts schema metadata when it does. Surfaced by PR #1930 (PAN-1847).

eltmon:
Closes [PAN-1932](https://github.com/eltmon/panopticon-cli/issues/1932)

When a database created by a newer build (higher `user_version`) was opened by older code, the migration guard `currentVersion === SCHEMA_VERSION` was false, no migration branch matched, and the trailing `user_version = SCHEMA_VERSION` pragma silently downgraded the schema-version metadata.

Changes:
- Change the guard in `src/lib/database/schema.ts` to `currentVersion >= SCHEMA_VERSION`.
- Regenerate `sync-sources/hooks/record-cost-event.js` from source (now includes the `>=` guard).
- Add `src/lib/database/__tests__/schema-migration-guard.test.ts` covering newer, exact, and older `user_version` cases.

Quality gates: targeted tests + full typecheck pass.


eltmon:

PR #1941, opened 2026-06-17
fix(config): remove hardcoded default conversation model ladder (PAN-1927)

### LINKED ISSUE #1927: fix(config): remove hardcoded model fallbacks — default/role model must come from explicit settings
## Principle (operator directive)

**No model fallback may be hard-coded. The default/fallback model must always come from an explicit setting that justifies it.** A hardcoded fallback can silently run a model the operator never chose and never sanctioned.

## The bug that surfaced it

`getDefaultConversationModelApi()` (`src/lib/settings-api.ts:273-285`) hard-codes a provider-ordered ladder of model IDs when `default_conversation_model` is unset or a role model fails to resolve:

```ts
if (config.defaultConversationModel) return resolveModelIdSync(config.defaultConversationModel);
if (config.enabledProviders.has('openai'))  return resolveModelIdSync('gpt-5.5');
if (config.enabledProviders.has('minimax')) return resolveModelIdSync('minimax-m2.7-highspeed');
if (config.enabledProviders.has('google'))  return resolveModelIdSync('gemini-3.1-pro-preview');
if (config.enabledProviders.has('kimi'))    return resolveModelIdSync('kimi-k2.5');
// ... zai, mimo, nous, dashscope — all hardcoded model IDs
```

**Observed symptom:** the PAN-1847 **planning** agent ran on **`gpt-5.5` (footer: "gpt-5.5 default")** even though `config.yaml` sets `roles.plan.model = workhorse:expensive = claude-opus-4-8` and Anthropic is enabled. The plan-role model was **not applied** — resolution fell through to the hardcoded `gpt-5.5` default.

## Scope

1. **Remove the hardcoded model-ID ladder** in `getDefaultConversationModelApi` (`settings-api.ts:278-285`). The default conversation model must come **only** from the explicit `default_conversation_model` setting. If it is unset → surface a clear, actionable error ("no default model configured — set models.default_conversation_model"), never a silent hardcoded pick.
2. **Trace + fix the role fall-through:** find why the `plan` role resolution dropped to the conversation default instead of `roles.plan.model` (`workhorse:expensive`) and fix it so a configured role model is always honored (or fails loudly).
3. **Audit every other hardcoded model fallback** — `resolveModelIdSync('<literal>')`, `model-fallback.ts` FALLBACK_MAP, role/workhorse defaults — and route each through an explicit setting. Any remaining default must trace to config, not a code literal.

## Acceptance
- No code path returns a hardcoded model ID as a default/fallback.
- An unset `default_conversation_model` produces an explicit error, not a silent model choice.
- A role with a configured `roles.<role>.model` always resolves to it (verified by a test for the `plan` role specifically).

Related state-model epic #1919 (state/config must be explicit, not implicit).

eltmon:
Closes [PAN-1927](https://github.com/eltmon/panopticon-cli/issues/1927)

 previously returned a provider-ordered ladder of hardcoded model IDs when  was unset. This violated the operator directive that no model fallback may be hardcoded.

Changes:
-  now returns  when no default is configured, and only returns the explicit setting when present.
- Added  for callers that must have a concrete default; it throws  when unset.
-  exposes  as optional so the rest of the settings surface does not break when no default is set.
- Updated unit and integration tests to assert the new explicit-only semantics.

Quality gates pass (targeted tests + typecheck).

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed implicit default conversation model fallback behavior. The application now requires explicit configuration and will fail with a clear error when no default conversation model is set, preventing unexpected behavior from undocumented defaults.

### DIFF
--- a/docs/overdeck-remodel/investigations/pipeline-transitions.md
+++ b/docs/overdeck-remodel/investigations/pipeline-transitions.md
@@ -49,6 +49,27 @@ todo → planning → planned/proposed → working → in-review → testing
                            · deacon-ignored · cancelled
 ```
 
+### 1a-def. Spine stages — definition + entry/exit conditions
+
+The ten happy-path stages, each as the operator means them. "Entry/exit" name
+the *logical* condition; the physical writes that realize them are in §2.
+
+| Stage | One-line meaning | Entry condition | Exit condition |
+|---|---|---|---|
+| **todo** | Issue exists, no plan, no work | Issue opened / reset / wiped | Planning starts |
+| **planning** | Planning agent is producing a vBRIEF | `start-planning` spawns plan role (`IssueState=in_planning`) | Planning completes or aborts |
+| **planned/proposed** | Plan exists, awaiting start | `complete-planning` writes `plan.status=proposed` + beads | `pan start` (or auto-start) |
+| **working** | A work agent is implementing beads | `plan.status=running`, `IssueState=in_progress`, work agent running | Work agent signals done |
+| **in-review** | Code review role is judging the diff | `work.completed` event → review role dispatched (`reviewStatus=reviewing`) | Review verdict (approved / blocked) |
+| **testing** | Test role runs automated + UAT verification | `review.approved` event → test role (`testStatus=testing`) | Test verdict (passed / failed / skipped) |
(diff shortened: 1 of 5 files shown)


eltmon:

PR #1943, opened 2026-06-17
fix(dashboard): persist pi-harness cost events into cost_events (PAN-1935)

eltmon:
## Problem

Pi-harness work agents (e.g. `kimi-k2.7-code`) were recording **zero cost** in Panopticon. The pi-extension already emits `cost-event` runtime events via the agent heartbeat endpoint, but the dashboard only translated them into `cost.event_recorded` domain events — it never wrote them to the canonical `cost_events` store or updated `agents.cost_so_far`.

Claude-code agents avoid this gap by writing cost events directly through the `record-cost-event.js` PostToolUse hook. Pi agents have no such hook, so their spend was invisible.

## Changes

- **Contracts**: extend `CostEventRecordedEvent` payload with optional `cacheReadTokens`, `cacheWriteTokens`, `model`, and `sessionType`.
- **Heartbeat translation** (`agent-event-utils.ts`): preserve those fields from the pi `cost-event` body, including `agentRole` → `sessionType` and a `costUsd`/`cost` fallback.
- **Cost projection** (`cost-event-projection.ts`): new subscriber that writes `cost.event_recorded` domain events to the canonical JSONL + SQLite `cost_events` table and atomically increments `agents.cost_so_far`.
- **Dashboard wiring** (`domain-services.ts`): subscribe the projection to the event store.
- **Pricing**: add `kimi-k2.7-code` to the Moonshot/Kimi pricing table so future token-based cost calculations resolve correctly.
- **Tests**: cover body translation, schema decoding, SQLite persistence, JSONL logging, and `cost_so_far` accumulation.

## Acceptance

- [x] Pi/kimi work agents now produce non-zero `cost_events` rows.
- [x] `agents.cost_so_far` increments as cost events arrive.
- [x] Dashboard cost surfaces can aggregate pi/kimi spend.

Burn-rate pause/escalation is intentionally left as a follow-up; once cost is visible, that threshold logic can be wired against `agents.cost_so_far` / `cost_events` in a subsequent change.

Closes PAN-1935

### DIFF
--- a/docs/overdeck-remodel/END-STATE.md
+++ b/docs/overdeck-remodel/END-STATE.md
@@ -1,0 +1,295 @@
+# Overdeck — End-State Architecture (DRAFT)
+
+> **Status:** working draft. The architecture and the two fully-audited domains
+> (Issues, Agents) are firm. Sections marked ⏳ are being filled from in-flight
+> current-state audits; Effect code blocks are marked `‹fill from
+> ARCHITECTURE-CONVENTIONS.md›` until the Effect v4-beta idiom audit lands.
+> Companion docs: [`ARCHITECTURE-CONVENTIONS.md`](ARCHITECTURE-CONVENTIONS.md)
+> (the Effect house style) and [`investigations/`](investigations/) (the
+> evidence base).
+
+## 0. The thesis
+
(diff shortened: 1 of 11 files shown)


eltmon:

PR #1945, opened 2026-06-17
fix(workspace): remove git rebase from auto-commit to protect shared main worktree (PAN-1929)

eltmon:
## Problem

`src/lib/pan-dir/auto-commit.ts` was running `git rebase origin/main` at the end of every background state/beads auto-commit. That is a history-rewriting operation in the **shared primary `main` worktree**, driven by a background process. It currently fails safely only because the beads janitor leaves the working tree dirty, but if the tree were clean and behind origin, the rebase would move shared HEAD and rewrite history while operator/conversation/agent work is live.

## Fix

Remove the `git rebase origin/main` block entirely. Auto-commits now create a local commit on `main` and stop. Records still reach git durably; the shared worktree is never rebased, never requires a clean tree, and never has its HEAD moved by a background process.

This is the minimal safety fix. Routing records to a dedicated infra repo (the `repoRoot` option already exists) and plumbing-based commits remain future improvements, but they are not required to eliminate the immediate hazard.

## Changes

- [`src/lib/pan-dir/auto-commit.ts`](): removed the post-commit `git rebase origin/main` and its abort fallback; added a PAN-1929 comment explaining why rebase is forbidden.
- [`src/lib/pan-dir/__tests__/auto-commit.test.ts`](): added a regression test that sets `origin/main` ahead of local `main`, runs an auto-commit, and asserts the new commit is a direct descendant of the previous local HEAD (no rebase / no move to origin's tip).

## Acceptance

- [x] No `git rebase` or history-rewriting operation runs in the shared primary worktree.
- [x] Auto-commit no longer requires a clean tree or moves shared HEAD.
- [x] Records still commit durably to git.
- [x] "rebase failed for main" log noise is eliminated.

Closes PAN-1929


eltmon:

PR #1946, opened 2026-06-17
fix(dashboard): source /api/health/agents membership from agents table (PAN-1914)

eltmon:
## Problem

`GET /api/health/agents` in [`src/dashboard/server/routes/misc.ts`]() was discovering agent membership by reading `~/.panopticon/agents` with `readdir()`. PAN-1908 moved the authoritative agent registry to SQLite, but this endpoint still scanned the filesystem, which is slower and can include stale/orphaned directories that the DB no longer tracks.

## Changes

- Import `listAllAgents` from the SQLite agents registry.
- Replace the `readdir(agentsDir)` + name-prefix filter with `listAllAgents()`.
- Extract the response-building logic into an exported `buildHealthAgentsResponse()` helper so it is unit-testable without spinning up the full HTTP router.
- Keep per-agent file reads (`state.json`, `health.json`, `context-pct`) as targeted enrichment for agents that are already known to the registry.
- Add [`src/dashboard/server/routes/__tests__/health-agents.test.ts`]() with regression tests covering:
  - membership sourced from the agents table
  - filtering of agents that `determineHealthStatus` hides
  - enrichment from `health.json` and `context-pct`

## Acceptance

- [x] `/api/health/agents` discovers agents from the agents table, not `readdir(agentsDir)`.
- [x] Per-agent files are read only for enrichment, not membership.
- [x] Existing behavior for liveness, health status, and context percent is preserved.

Closes PAN-1914


eltmon:

Issue #2248, opened 2026-07-02
bug(test): deacon-swarm-synthesis.test.ts is order-dependent — passes isolated, fails intermittently in full-suite CI


eltmon:
Evidence (2026-07-02 evening): three tests in tests/unit/lib/cloister/deacon-swarm-synthesis.test.ts ('dispatches a synthesis-phase slot...', 'dispatches implementation with synthesis context...', 'dispatches single-parent items directly...') fail with 'AssertionError: expected [ Array(1) ] to deeply equal [ Array(1) ]' in full-suite CI runs, but pass in isolation (npx vitest run <file> → 3 passed, verified locally on main @ 0b46ac7111).

Occurrences: main run on 1e82badc32 FAILED 21:17Z; main run on 0b46ac7111 PASSED 21:25Z; feature/pan-2156 and strike/pan-1935 branch runs both FAILED on exactly this file (~21:33Z, ~22:26Z). Disjoint branches + intermittent main = state leakage between tests under parallel workers, not branch code.

Two investigation leads: (1) the deep-equal arrays differ in some field only when other tests precede — dump the actual diff; (2) timing-wise the flake appears right after 1e82badc32 (mechanical PRD-first gate, which touches pan-dir/drafts and issues routes) — check whether a stray drafts dir from another test changes slot dispatch content. Repro guidance: npx vitest run tests/unit/lib/cloister/ (the directory together) or the full suite; NOT the file alone. Per repo rules: fix isolation (fresh tmp dirs / reset module state in beforeEach), never serialize with maxForks=1.

eltmon:
Fixed and merged: PR #2264 (squash) isolates the three order-dependent dispatch tests in deacon-swarm-synthesis.test.ts — tests-only diff, no assertions weakened, full-suite CI green on the PR. This flake red-mained main once and blocked two PRs on 2026-07-02; branch runs should stop coin-flipping now. Fix authored by the strike agent; PR创建 and merge shepherded by the orchestrator after the agent idled post-push.

PR #2256, opened 2026-07-03
PAN-2248

eltmon:
**Issue:** #2248


eltmon:
Closing as stale duplicate — commits already patch-equivalent on origin/main (verified via git cherry). Issue PAN-2248 landed via another path.

Issue #2258, opened 2026-07-03
bug(dashboard): strike agents invisible in the issue tree — contracts event-reducer layer has no strike concept (PAN-1506 class)


eltmon:
Reported by operator 2026-07-02: strike-pan-2248 (live, working) does not appear under PAN-2248 in the Command Deck issue tree.

Verified seam-by-seam (all server-side layers are strike-aware; the drop is in the shared store layer):
- /api/agents returns strike-pan-2248 (issueId PAN-2248, role strike) OK
- resource-discovery.ts includes 'strike-' in DISCOVERABLE_SESSION_PREFIXES (PAN-1682) and attributes the session — the PAN-2248 tree node carries agentStatus: 'active' OK
- read-model.ts VALID_ROLES includes 'strike' OK
- Frontend components fully support strikes (SessionNode.tsx strike icon/case, RoleBadge.tsx strike style) OK
- BUT: zero occurrences of 'strike' in packages/contracts/src/event-reducers.ts or events.ts — the domain-event/store layer that feeds the tree's per-issue session chips has no strike concept, so the chip never materializes even though the node shows an active status.

History: PAN-1506 ('strike agents missing from frontend store despite appearing in /api/agents and read-model bootstrap') is CLOSED — this is either its regression or a layer its fix never covered (store ingestion vs contracts reducers). The fix should include a regression test at the reducer level: given an agent domain event with role 'strike', the reduced store must yield a session under the issue (that's the mechanical gate PAN-1506 lacked, which is why this resurfaced).

Not pipeline machinery — safe for normal pipeline flow.

eltmon:
Fixed and merged via PR #2271: contracts event-reducers now project role=strike agent events into per-issue sessions, with the reducer-level regression test that PAN-1506 lacked. Strikes will appear in the Command Deck issue tree from the next dashboard deploy/reload onward.

PR #2269, opened 2026-07-03
PAN-2258

eltmon:
**Issue:** #2258


eltmon:
Closing as stale duplicate — commits already patch-equivalent on origin/main (verified via git cherry). Issue PAN-2258 landed via another path.

Issue #1198, opened 2026-05-18
Workspace init container's bun install doesn't populate container-node-modules named volume


eltmon:
## Problem

The workspace devcontainer compose template defines a named volume `container-node-modules` that shadows `/workspaces/panopticon/node_modules` for the `init`, `frontend`, and `server` services. The init container's job is to populate it by running:

```
bun install --filter '@panctl/cli' --filter '@panctl/contracts' --filter '@panopticon/pi-extension' --filter 'panopticon-server' --filter 'panopticon-dashboard'
```

But the install never lands in the named volume. Direct inspection after init exits 0:

```
docker run --rm -v panopticon-feature-pan-1052_container-node-modules:/v alpine ls /v | wc -l
0
```

…and the server service then fails at startup with `Cannot find package 'effect' imported from /workspaces/panopticon/dist/dashboard/server.js` because there are no packages at the runtime resolution path.

## Reproduction (2026-05-18)

1. `pan workspace rebuild PAN-1052` (after `pan workspace render-devcontainer feature-pan-1052` — i.e., with the latest post-PAN-1169 template).
2. Init logs report `1230 packages installed` including `effect@4.0.0-beta.45` — so install ran.
3. `pan workspace ps` shows `server-1 service exited (1)`.
4. `docker volume inspect panopticon-feature-pan-1052_container-node-modules` exists; `docker run -v <vol>:/v ... ls /v` returns 0 entries.

## Hypothesis

The init container's compose volume declaration may not be effective at install time, OR `bun install --filter` is writing somewhere other than `/workspaces/panopticon/node_modules`. The `dev` service (which has only the bind mount, no named-volume shadow) sees all 33 expected top-level entries via the bind mount — so a host `bun install` works. But init writes don't reach the named volume.

Also worth checking: the `bun-store` named volume gets the same treatment and was also empty (`0 entries`).

## Workaround tonight

For PAN-1052: copied the host's working `node_modules` directly into the named volume, then restarted server-1.

```bash
docker run --rm \
  -v <workspace>/node_modules:/src:ro \
  -v panopticon-feature-pan-1052_container-node-modules:/dst \
  alpine cp -a /src/. /dst/
```

Not a real fix — but unblocks resume in one shot when the host install already happened.

## Related

- PAN-1169 (bun --filter top-level node_modules fix) — already merged but doesn't address this scenario.
[...]

PR #2778, opened 2026-07-16
fix(infra): repair devcontainer node_modules volume ownership (PAN-2777)

eltmon:
Workspace init failed with ~1,618 `failed to link package` errors because the `container-node-modules` named volume mounts **root-owned** while init runs as `node` (uid 1000) — proven in-container: `touch node_modules/.writetest` → Permission denied. Everything downstream (bun link ENOENT/EPERM, unlinked husky, exit 127) was ownership. PAN-2770's `--ignore-scripts` was necessary but insufficient.

## Change

- New `init-perms` service: runs as root, `chown node:node /workspaces/overdeck/node_modules`, ordered before init via `depends_on: service_completed_successfully`. Runs on **every** init, so it also heals volumes created root-owned before this fix.
- First (disproven) attempt `--backend=copyfile` reverted — the final template keeps only `--ignore-scripts`.
- Read-only mount of the primary checkout's `.git` at the worktree pointer path, so in-container builds can read the current commit.
- Template test asserts the perms step exists, runs as root, and orders before install.

## Verification (flywheel orchestrator, live)

- Root cause proven in-container (root-owned mountpoint, write denied as node).
- Heal verified live on `feature-pan-2168`: volume now `node:node`, writable.
- **Full init run: exit 0** (was exit 1 with 1,618 failures).
- Template test 3/3, typecheck exit 0.

Unblocks PAN-2168, PAN-2255, PAN-2532 (with a `pan workspace render-devcontainer` re-render each). Related: #1198.

Closes PAN-2777

### DIFF
--- a/infra/.devcontainer-template/docker-compose.devcontainer.yml.template
+++ b/infra/.devcontainer-template/docker-compose.devcontainer.yml.template
@@ -19,6 +19,17 @@ services:
       - devnet
       - overdeck
 
+  # Docker creates named-volume roots as root. Repair ownership on every init so
+  # this also heals volumes created before PAN-2777.
+  init-perms:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    user: root
+    command: chown node:node /workspaces/overdeck/node_modules
+    volumes:
+      - container-node-modules:/workspaces/overdeck/node_modules
+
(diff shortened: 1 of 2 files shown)


eltmon:
Closed via close-out ceremony

PR #2826, opened 2026-07-17
fix(cloister): authenticate the automatic work-agent spawn (PAN-2825)

### LINKED ISSUE #2825: plan→work autoSpawn silently 401s: complete-planning POSTs /api/agents without the internal token (regression from 4697f17630)
**The plan→work handoff has been silently dead since a build containing 4697f17630 went live (~2026-07-16 midday).** Every `pan start <id>` on an unplanned issue promises "the work agent will start automatically after planning finalizes." Planning finalizes, writes a good spec — and no work agent ever starts. No error surfaces.

Observed on PAN-2822 and PAN-2661 this morning; retroactively on PAN-1897, PAN-2045, PAN-2252, PAN-2807, PAN-2647.

## Causal chain (verified end to end)

**1. The guard landed.** `4697f17630` (2026-07-16 00:25, "unify the three issue views into one progressive-density Issue View", PAN-2499) added to `src/dashboard/server/routes/agents/spawn.ts:189` on `POST /api/agents`:

```ts
const authError = rejectUnsafeDashboardMutationRequest(request);
if (authError) return authError;
```

This is correct hardening and should stay.

**2. The server's own internal caller sends no credentials.** `completePlanningAutoSpawn()` in `src/lib/overdeck/planning-promotion.ts:292-299`:

```ts
const response = await (options.fetchImpl ?? fetch)(new URL('/api/agents', dashboardOrigin), {
  method: 'POST',
  headers: {
    'content-type': 'application/json',
    origin: dashboardOrigin,          // <-- no INTERNAL_TOKEN_HEADER, no session cookie
  },
  body: JSON.stringify({ issueId: options.issueId, role: 'work' }),
});
```

`rejectUnsafeDashboardMutationRequest` → `rejectUnauthorizedDashboardRequest` (`dashboard-auth.ts:303-314`) → `hasDashboardAuth()`, which accepts **only** a valid `x-overdeck-internal-token` / `Authorization: Bearer` (`hasDashboardInternalTokenHeaders`, line 103) or the `DASHBOARD_SESSION_COOKIE`. The POST carries none → **HTTP 401 `{"error":"unauthorized"}`**.

**3. The recovery path can't catch it.** `classifyAutoSpawnSkip()` (`planning-promotion.ts:269-276`) checks for `stackHealth`, `paused`, `troubled`, `guardrails`/409 — a 401 with `{error:'unauthorized'}` matches none, so it returns `'spawn-failed'`. The `stack-unhealthy` → `rebuild-and-start` recovery at line 317 never fires.

**4. It reports success.** The terminal phase emits `status: 'success'` while carrying `workAgentSpawned: false`:

```
phase=autoSpawn status=start   reason="posting work-agent spawn request" dashboardOrigin="http://127.0.0.1:3011"
phase=autoSpawn status=skipped reason="spawn-failed" httpStatus=401 error="unauthorized"
phase=terminal  status=success reason="spawn-failed" autoSpawn=true workAgentSpawned=false
```

Nothing alerts, nothing retries. The issue sits at `plan.status: proposed` with a finalized spec and no agent, indefinitely.

## Why it surfaced when it did

`autoSpawn` phase events show a hard boundary: last `success` at **2026-07-16T05:06** (PAN-2772/2773, via the stack-unhealthy→rebuild recovery), then `spawn-failed` on **every** attempt from 17:53 onward. The commit merged at 00:25; the failure begins when a build containing it was deployed. `git merge-base --is-ancestor 4697f17630 57222e4bcb` confirms it is in the live build.

It stayed invisible because the Flywheel drives issues with an explicit `pan start` on already-planned issues — the **CLI** path resolves the internal token and authenticates, so operator/orchestrator-driven work kept spawning normally. Only the *automatic* plan→work handoff is broken, which is exactly the path an unattended run depends on.

## Fix

`completePlanningAutoSpawn` runs inside the dashboard server process and already imports from the same lib tree as `getInternalTokenSync()`. Attach the internal token to the spawn POST (and to the `rebuild-and-start` recovery POST at line 318, which has the same omission):

```ts
import { getInternalTokenSync, INTERNAL_TOKEN_HEADER } from '../internal-token.js';
...
const internalToken = getInternalTokenSync();
headers: {
  'content-type': 'application/json',
  origin: dashboardOrigin,
  ...(internalToken ? { [INTERNAL_TOKEN_HEADER]: internalToken } : {}),
},
```

Two hardening items in the same pass, both of which would have made this loud instead of silent:

- **`classifyAutoSpawnSkip` should classify 401/403 as its own `unauthorized` reason** rather than folding them into generic `spawn-failed`.
- **The terminal phase must not report `status: 'success'` when `workAgentSpawned === false`.** A requested spawn that did not happen is a failure; emitting success is what hid this for 16 hours.

## Acceptance

- `pan start <fresh-planned-issue>` auto-starts the work agent with no manual `pan start` follow-up.
- An autoSpawn phase event shows `status=success` with a real `agentId`.
- A unit test asserts the spawn POST carries `INTERNAL_TOKEN_HEADER` and that a 401 response yields a non-success terminal phase.

Grep for other in-process callers that POST authenticated dashboard routes without the token — this is unlikely to be the only one.

eltmon:
Fixes #2825 — the plan→work autoSpawn handoff has been silently 401ing since a build containing 4697f17630 went live (~2026-07-16 midday).

## What was broken

`4697f17630` correctly added `rejectUnsafeDashboardMutationRequest` to `POST /api/agents`. But the server's own in-process callers post to that route with no credentials, so they got 401 `{"error":"unauthorized"}`. The skip classifier didn't recognize 401, folded it into generic `spawn-failed`, and the terminal phase then reported `status: 'success'` while carrying `workAgentSpawned: false` — so 16 hours of dead handoffs made no noise.

## What this changes

- **Authenticate both in-process spawn callers.** `completePlanningAutoSpawn` and the `rebuild-and-start` recovery POST now attach `INTERNAL_TOKEN_HEADER`. A second caller with the identical defect — `spawnWorkAgentThroughAgentsEndpoint` in `orphan-proposed-reconciler.ts` — is fixed the same way. That one matters: it is the backstop that recovers specs stranded at `proposed`, so it was 401ing exactly when it was needed.
- **Classify 401/403 as `unauthorized`** instead of generic `spawn-failed`, in both classifiers.
- **Stop reporting success on a failed spawn.** `resolveCompletePlanningTerminalStatus` emits `failure` when autoSpawn was requested but no agent started.

The auth guard stays — this fixes the callers, not the guard.

## Verification

- `complete-planning` — 15/15 passing (new coverage asserts the spawn POST carries the internal token and that a 401 yields a non-success terminal phase)
- `orphan-proposed-reconciler` — 27/27 passing
- typecheck clean (no new errors); lint passed

Landed as an operator-authorized strike: diff reviewed and gates re-run independently by the Flywheel orchestrator (RUN-65) before merge.

### DIFF
--- a/src/dashboard/server/routes/__tests__/complete-planning.test.ts
+++ b/src/dashboard/server/routes/__tests__/complete-planning.test.ts
@@ -1,8 +1,19 @@
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import { existsSync, mkdirSync, mkdtempSync, readFileSync, readdirSync, rmSync, writeFileSync } from 'node:fs';
 import { mkdir } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
+
+const internalTokenMocks = vi.hoisted(() => ({
+  getInternalTokenSync: vi.fn(() => 'test-internal-token'),
+}));
+
+vi.mock('../../../../lib/internal-token.js', async (importOriginal) => ({
+  ...await importOriginal<typeof import('../../../../lib/internal-token.js')>(),
+  getInternalTokenSync: internalTokenMocks.getInternalTokenSync,
+}));
+
+import { INTERNAL_TOKEN_HEADER } from '../../../../lib/internal-token.js';
 import {
   beginCompletePlanningLease,
   completePlanningArtifacts,
@@ -10,6 +21,7 @@ import {
   completePlanningAutoSpawnAndKill,
   completePlanningFilesToStage,
   completePlanningWorkspaceGitAddCommands,
+  resolveCompletePlanningTerminalStatus,
 } from '../../../../lib/overdeck/planning-promotion.js';
 import { applyStatusOverrides } from '../../../../lib/vbrief/io.js';
 import { lintPlanQuality, PlanQualityLintError } from '../../../../lib/vbrief/quality-lint.js';
@@ -384,7 +396,10 @@ describe('completePlanningArtifacts', () => {
     const fetchImpl: typeof fetch = async (input, init) => {
       expect(String(input)).toBe('http://127.0.0.1:3011/api/agents');
       expect(init?.method).toBe('POST');
-      expect(init?.headers).toMatchObject({ origin: 'http://127.0.0.1:3011' });
+      expect(init?.headers).toMatchObject({
+        origin: 'http://127.0.0.1:3011',
+        [INTERNAL_TOKEN_HEADER]: 'test-internal-token',
+      });
       expect(JSON.parse(String(init?.body))).toEqual({ issueId: 'PAN-1146', role: 'work' });
       return new Response(JSON.stringify({ success: true, agentId: 'agent-pan-1146' }), { status: 200 });
     };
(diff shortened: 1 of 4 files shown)


eltmon:
## Merge Failed

Work agent strike-pan-2825 stopped before completing the rebase onto main

PR #3091, opened 2026-07-26
fix(test): pin forwarded inspect notes arg in itemId attribution test (PAN-3088)

### LINKED ISSUE #3088: RED MAIN: PAN-3077/3078 (1302ccf6d2) regressed inspect itemId attribution — checkpoint receives notes prose instead of the structured itemId
## Summary

`main` is RED. Last green is `cce23592b8`; the break is `1302ccf6d2` — "fix(pipeline): deliver inspect verdicts and always pass --effort (PAN-3077, PAN-3078)".

Failing: `tests/unit/dashboard/server/routes/specialists-inspect-item.test.ts` > **"checkpoints the exact structured itemId regardless of notes wording"** (1 file failed, 1162 passed).

Every feature PR now inherits the failing `test` check, so nothing reaches the merge gate until this is fixed.

## The failure

```
AssertionError: expected "vi.fn()" to be called with arguments: [ 'overdeck', 'PAN-2724', …(3) ]
+   "This predates this bead and is correct",
```

The checkpoint call received **prose from the verdict notes** where the **structured itemId** (`PAN-2724`) was expected. That is precisely the regression the test name guards against: item attribution must come from the structured field, never from parsing notes wording.

## Why this matters beyond a red test

This is not a stale-mock mismatch — it is the assertion catching a real behavioural change. Inspect verdicts are checkpointed **per item**; if the itemId is derived from notes text, checkpoints attach to the wrong item (or to a non-existent one) whenever a reviewer's prose happens to mention or omit an id. Wrong-item attribution is silent: the checkpoint succeeds, it just records against the wrong thing.

## Fix direction

Have the inspect-verdict delivery path read the itemId from the structured verdict field it already carries, rather than from notes. Keep the existing test green as the contract — it encodes exactly the invariant that broke. If the new `--effort` plumbing changed the shape of what reaches the checkpoint call, thread the structured id through rather than recovering it downstream.

Reproduce with:

```
npx vitest run tests/unit/dashboard/server/routes/specialists-inspect-item.test.ts
```

Flywheel-Affects-Criterion: 3

eltmon:
Fixes #3088. Restores `main` to green.

`main` went red on `tests/unit/dashboard/server/routes/specialists-inspect-item.test.ts` > "checkpoints the exact structured itemId regardless of notes wording" after `1302ccf6d2` (PAN-3077/PAN-3078).

**No production behaviour regressed.** `onInspectComplete` gained a sixth `notes` parameter so the verdict message delivered to the work agent can carry the blocking finding. `toHaveBeenCalledWith` requires an exact argument match, so the *added* argument failed the assertion. The structured itemId is still passed in its own position.

**The assertion is extended, not loosened.** It now pins both invariants: attribution comes from the structured field, and the prose travels as its own argument rather than leaking into the itemId. That is strictly stronger than the original test.

Note: the issue body's original root cause (written by me) claimed itemId attribution had regressed and warned of silent wrong-item checkpointing. That was a misdiagnosis — I read vitest's `+` line as a substituted value when it was an added argument. Corrected in a comment on the issue.

**Gate:** `npx vitest run tests/unit/dashboard/server/routes/specialists-inspect-item.test.ts` — 3 passed.

Filed, misdiagnosed, corrected and driven by the Flywheel (RUN-70).

https://claude.ai/code/session_01LmAFhdqW76Z5bnmkLRXkMH

### DIFF
--- a/tests/unit/dashboard/server/routes/specialists-inspect-item.test.ts
+++ b/tests/unit/dashboard/server/routes/specialists-inspect-item.test.ts
@@ -107,12 +107,16 @@ describe('POST /api/specialists/done inspect item attribution', () => {
     });
 
     expect(result.status).toBe(200);
+    // The itemId argument is the structured `itemId` field verbatim. The notes
+    // ride along as their own argument (PAN-3078 forwards them into the verdict
+    // message delivered to the work agent) and must never leak into attribution.
     expect(mocks.onInspectComplete).toHaveBeenCalledWith(
       'overdeck',
       'PAN-2724',
       'issue-view-model',
       'passed',
       join(projectPath, 'workspaces', 'feature-pan-2724'),
+      'This predates this bead and is correct',
     );
   });
 });


eltmon:
Superseded by #3087 (PAN-3086), which landed the same fix as 5f92731d7d before this branch could merge. Closing rather than resolving the conflict — the work is already on main.

Issue #3092, opened 2026-07-26
Review agent's own record lock blocks its own verdict write: specialists/done collides, fallback is consumed, reviewStatus stuck at reviewing (MIN-902)


eltmon:
## Summary

A review agent can hold the per-issue record lock in a way that blocks **its own** verdict write. The reviewer completes, reports APPROVED, and every `specialists/done` signal collides with the lock held by the reviewer's own session. The workspace fallback is consumed, and `reviewStatus` stays `reviewing` forever. The passed verdict is lost from every persisted plane.

## Observed 2026-07-26 00:09Z — MIN-902

`agent-min-902-review` reports the review APPROVED at `fe@5eca65e0 api@8083efe8` with its report written. Verified independently:

- SQLite `review_status`: `review_status: reviewing`, `test_status: pending` (updated 00:09:07Z)
- Durable record `min-902.json` `pipeline`: `reviewStatus: reviewing` — **both planes agree, and both are wrong**
- **No fallback file remains** at `.overdeck/pipeline-verdict.json*` — the fallback was written and consumed, so the PAN-2989 drain cannot recover it

So the verdict exists only in the reviewer's report artifact and its own claim. Nothing in the persisted state records that this review passed.

## Why the existing recovery paths do not apply

- **PAN-2989's verdict-fallback drain** is PID-aware by design: `claimWorkspaceVerdictFallbackSync` skips claims whose owning process is **alive** (`isPidAlive`), because a live pid means an in-flight drain owns that generation. Here the lock holder is a live, healthy reviewer, so the drain correctly declines — and the fallback has already been consumed anyway.
- **`pan review resync`** heals a lost pipeline *event* by re-emitting canonical status. I ran it: `✓ Re-synced MIN-902: review=reviewing test=pending`. It faithfully re-emitted the canonical value, which is exactly the stuck one. Resync cannot manufacture a verdict that never reached the write door.
- **`pan review restart`** would redispatch a review that is already complete and unchanged — burning a full review cycle to recover a verdict that already exists, and the reviewer explicitly asked that this not happen.

There is no sanctioned path for an orchestrator to persist an already-earned verdict. The write door is `specialists/done`, and that is the call being blocked.

## Root-cause question for whoever fixes this

[...]

PR #3222, opened 2026-07-28
PAN-3092

eltmon:
**Issue:** #3092

## Acceptance Criteria

- [ ] Verdict-aware pipeline merge: new pipeline-verdict-merge.ts, pickNewerPipeline delegates
- [ ] Drain supersede check becomes content-aware (no more unlanded fallback deletion)
- [ ] Bounded exponential backoff for terminal verdict journal writes + CLI do-not-retry message
- [ ] Deacon sweep patrol drains stranded verdict fallbacks and surfaces >10min contention
- [ ] Frozen test agent escalates to needs-you instead of 'none' forever
- [ ] Role prompts and test spawn prompt carry the do-not-retry stop-signal contract
- [ ] docs/AGENT-STATE-PLANES.md documents the new verdict-durability semantics

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved terminal review, UAT, and test verdicts during concurrent and newer updates.
  * Added verdict-conflict handling to avoid silently discarding contradictory terminal verdicts.
  * Improved recovery for stalled test agents and stranded verdict fallbacks, including clearer “do not re-run” messaging and automatic fallback draining.
* **Documentation**
  * Expanded guidance on verdict fallback behavior, merge rules, conflict outcomes, and operator recovery steps.
* **Tests**
  * Added coverage for verdict preservation, fallback draining under contention, conflict scenarios, and single-attempt verdict signaling.  
* **Chores**
  * Reduced duplicate activity/event emissions via at-most-once internal event append semantics.


eltmon:
# Review CHANGES REQUESTED for PAN-3092

# Review — PAN-3092 — 2026-07-28

## Verdict: CHANGES REQUESTED — the contention escalation uses the same contended record lock, and two terminal-verdict durability paths can still discard evidence

## Context
- Branch: `feature/pan-3092`
- HEAD reviewed: `566db378cb92103431348231524716843b9cf751`
- Manifest: `.pan/review/agent-pan-3092-review-566db378/context.json`
- Dimensions reviewed: correctness/logic, security, requirements/acceptance criteria, performance

## Blocking Findings

### ! [correctness / requirements] Long-contention surfacing is written through the lock that just failed — `src/lib/cloister/deacon-verdict-fallback-sweep.ts:85`
**Evidence tier:** Tier 1 — static
**Requirement:** FR-4 requires a fallback undrained for more than ten minutes to emit one activity warning and one needs-you trip. The PRD design explicitly says the activity entry is immediate and lock-independent, while the recovery trip is best-effort.
**Changed code:** After `drainWorkspaceVerdictFallback(issueId)` returns `false`, the sweep first awaits `recordRecoveryFailure(...)` at lines 85-87 and emits the activity entry only after that write succeeds at line 90. `recordRecoveryFailure()` writes the same per-issue record through `updateIssueRecordForWorkspace()` (`src/lib/cloister/recovery-trip.ts:14`). The outer catch at lines 100-102 silently absorbs any failure.
**Problem:** The condition being surfaced is sustained contention on that per-issue record lock. The needs-you write immediately attempts to acquire the same lock, so it can fail for exactly the reason the drain failed. When it does, control jumps to the outer catch before the lock-independent activity warning is emitted.
**Runtime impact:** During the exact >10-minute contention episode FR-4 targets, the patrol can emit neither the activity warning nor the needs-you trip. If the lock later frees, the next patrol drains the fallback and skips escalation, so the operator may never learn that the verdict was stranded.
**Fix:** Make the activity warning independent of `recordRecoveryFailure` and dedupe it without the record lock, then retry the durable recovery-trip write best-effort on later patrols. Do not place the only immediate operator signal after a write through the contended lock. Add a test where `drainWorkspaceVerdictFallback` returns false and `recordRecoveryFailure` rejects; the activity warning must still be emitted exactly once.

### ! [correctness / requirements] UAT verdicts are classified for backoff but are absent from every durable pipeline projection — `src/lib/pan-dir/pipeline-verdict-merge.ts:27`
**Evidence tier:** Tier 1 — static
**Requirement:** FR-3 requires NEW terminal verdict detection and durable fallback behavior for review, test, UAT, verification, inspect, and merge gates.
**Changed code:** `TERMINAL_GATE_VALUES` includes `uatStatus`, so `carriesNewTerminalVerdict()` gives UAT verdicts the new backoff. But `PanIssuePipelineRecord` has no `uatStatus`/`uatNotes` fields (`src/lib/pan-dir/record.ts:149-190`), `projectPipeline()` does not project them, and `durableSubset()` omits them (`src/lib/overdeck/review-status-record-sync.ts:161-198`). `pan admin specialists done uat` remains a reachable writer (`src/cli/commands/specialists/done.ts:212-220`), and UAT still participates in merge eligibility.
**Problem:** If a UAT terminal write exhausts the backoff, `writeWorkspaceVerdictFallbackSync()` writes a fallback that contains no UAT verdict or notes. The CLI can then state that the verdict is durable even though the fallback discarded the gate that triggered the durability path.
**Runtime impact:** A contended UAT pass/fail can still vanish from the journal, DB, and fallback, leaving merge eligibility stale and reproducing the verdict-loss class this issue is meant to eliminate.
**Fix:** Carry `uatStatus` and `uatNotes` through `PanIssuePipelineRecord`, `projectPipeline()`, `durableSubset()`, journal enrichment/reconciliation, and focused fallback/backoff tests. If UAT is intentionally retired, remove it from the terminal-verdict contract and the still-reachable writers instead of claiming it is durable.

### ! [correctness / requirements] The drain treats any terminal journal value as covering a different fallback verdict — `src/lib/pan-dir/pipeline-verdict-merge.ts:155`
**Evidence tier:** Tier 1 — static
**Requirement:** FR-2 states that supersede requires per-gate content equality; the PRD says any terminal-value mismatch returns `false`.
**Changed code:** `pipelineCoversFallbackVerdicts()` checks only whether the journal value is terminal. The new test explicitly declares a different terminal value covered.
**Problem:** `pipeline.updatedAt` is restamped by unrelated bookkeeping, which is the root defect documented by this issue. A concurrent writer can therefore carry an older terminal value into a later-timestamped journal while a newer specialist result is waiting in the fallback. Treating terminalness as proof of coverage reuses the same timestamp-conflation assumption the change is intended to remove.
**Runtime impact:** A fallback containing `passed` can be deleted when the journal contains a carried-forward `blocked`/`failed` value with a newer whole-pipeline timestamp, losing the only written copy of the newer result. This also directly contradicts the accepted FR-2 contract.
**Fix:** Require the journal gate value to equal the fallback gate value, as specified, and make the fold/merge path preserve the fallback until there is positive per-gate evidence of a newer conflicting terminal verdict or a strictly newer review cycle. Add the concurrent stale-terminal/bookkeeping-write regression case rather than asserting that any terminal value covers the fallback.

## Non-blocking Findings

### ~ [performance] Frozen-test escalation re-enters the nudge cycle every 30 minutes — `src/lib/cloister/deacon.ts:1417`
`alreadyNudged` is only true inside `NUDGE_DEDUP_MS`. Escalation does not persist a “do not nudge this dispatch again” marker, so after 30 minutes the same live idle generation returns to `nudge-write`; the following patrol escalates again but the recovery trip suppresses only the needs-you emission. This can keep waking an agent after the operator has already been notified. Track the nudge/escalation by dispatch generation, or suppress further delivery while that recovery trip remains open.

## Security
No blocking security findings. The change adds no new command execution from untrusted input, authentication bypass, secret handling, or destructive request path. Lock-owner text is rendered as log data rather than executed.

## Performance
No additional blocking performance finding beyond the operator-signal failure above. The verdict wait is asynchronous and bounded by the configured schedule, retry tests use fake timers, and the patrol uses asynchronous fallback reads. The sequential per-issue filesystem scan is bounded by the in-pipeline issue set and performs record writes only when a fallback exists.

## Requirements Covered
- FR-1: verdict-aware same-cycle merge and companion-field movement are implemented and tested.
- FR-3: bounded terminal-verdict backoff, CLI flush, fallback warning, and fake-timer tests are present for the gates actually projected durably.
- FR-5: the patrol is registered and attempts to drain live fallback files and stranded claims.
- FR-6: the frozen test decision now escalates without fabricating a verdict.
- FR-7: review/test prompt surfaces carry the one-signal contract.
- FR-8: the state-plane documentation was updated.
- NFR-1/NFR-2/NFR-3/NFR-4: the record lock protocol and durability budget are unchanged; delay tests use fake timers; recovery does not invent verdicts; no synchronous server I/O was added.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3092/.pan/review/agent-pan-3092-review-566db378/review.md

## Required action

`pan review request PAN-3092 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-3092

# Review — PAN-3092 — 2026-07-28

## Verdict: CHANGES REQUESTED — conflicting verdicts are preserved but surfaced as transient lock contention, and the “one warning per episode” dedupe is lost on restart

## Context
- Branch: `feature/pan-3092`
- HEAD reviewed: `981f88bcf04b49efad757967577a6955c3de2f65`
- Prior reviewed HEAD: `566db378cb92103431348231524716843b9cf751`
- Cycle scope: rework commit `d49217ad6596e16c6810e4ec8b85ed14109502ab`; merged `origin/main` changes are out of PR scope
- Manifest: `.pan/review/agent-pan-3092-review-981f88bc/context.json`
- Dimensions reviewed: correctness/logic, security, requirements/acceptance criteria, performance

## Blocking Findings

### ! [correctness / requirements] A terminal-verdict conflict is reported as a lock that will clear automatically — `src/lib/cloister/deacon-verdict-fallback-sweep.ts:95`
**Evidence tier:** Tier 1 — static
**Requirement:** The rework and updated documentation say that when journal and fallback hold different terminal values, the drain preserves both and the sweep puts the conflict in front of the operator rather than guessing a winner.
**Changed code:** `drainWorkspaceVerdictFallback()` now deliberately returns `false` for `pipelineConflictsWithFallbackVerdicts()` (`src/lib/overdeck/review-status-record-sync.ts:461-476`). The sweep treats every `false` result as record-lock contention and emits: “the per-issue record lock is contended … folds automatically once the lock frees,” using recovery path `verdict-fallback-contention`.
**Problem:** `false` now has two materially different meanings: a fold failed because the lock is contended, or the fold was intentionally withheld because two terminal verdicts conflict. In the conflict case there may be no lock owner at all, and freeing the lock cannot resolve the fallback because every later drain detects the same conflict and withholds the fold again.
**Runtime impact:** The operator receives false remediation and can wait indefinitely for an automatic fold that is structurally forbidden. The issue remains stuck with two written verdicts while the needs-you surface calls it transient contention instead of naming the conflicting gate/values and asking for adjudication.
**Fix:** Return a discriminated drain outcome (for example `drained | contended | verdict-conflict`) or otherwise expose the conflict reason to the sweep. Use a distinct `verdict-fallback-conflict` recovery path and an actionable message naming the gate and both terminal values; state that an operator must select the valid verdict or start a newer review cycle. Keep the contention message only for actual write failure.

### ! [requirements] The activity-warning dedupe is process-local, so a dashboard restart repeats the same episode — `src/lib/cloister/deacon-verdict-fallback-sweep.ts:39`
**Evidence tier:** Tier 1 — static
**Requirement:** FR-4 requires one activity-log warning and one needs-you trip per contention episode; NFR-5 requires episode-level dedupe.
**Changed code:** `warnedContentionEpisodes` is an in-memory `Set`. The warning is emitted before `recordRecoveryFailure()` is called, so after a dashboard restart the empty Set causes the warning to be emitted again even when the durable recovery trip already exists and `recordRecoveryFailure()` subsequently returns `emitNeedsYou: false`.
**Problem:** The durable trip is described as the cross-restart dedupe, but the code consults it only after emitting the warning. It therefore cannot dedupe the warning across restarts. If the trip write was previously blocked, the same duplicate also occurs after restart because neither dedupe plane survives.
**Runtime impact:** One prolonged fallback episode produces a new operator warning after every dashboard restart, violating the named no-nag-loop requirement. The Set also grows for the lifetime of the server because completed episode keys are never removed.
**Fix:** Give the activity warning a durable idempotency key or consult the canonical recovery-trip state before emitting it, while preserving the lock-independent first-warning guarantee when no trip can be recorded. Add a restart simulation that clears module-local state while retaining an open trip and proves the same fallback generation does not emit a second activity entry.

## Non-blocking Findings

### ~ [test coverage] The frozen-agent regression test does not advance beyond the 30-minute nudge window — `tests/lib/cloister/deacon-test-signal.test.ts:250`
The new Set makes the production behavior correct, but the third patrol runs immediately after the second despite its comment claiming coverage after the dedup window reopens. Advance fake time by more than `NUDGE_DEDUP_MS` before the third pass so the test fails if generation-level escalation suppression is later removed.

## Previous Findings Rechecked
- **Lock-independent immediate warning:** Fixed for a single server process; `recordRecoveryFailure()` failure no longer suppresses the activity entry.
- **UAT durability:** Fixed through `PanIssuePipelineRecord`, `projectPipeline()`, `durableSubset()`, note enrichment, terminal-value correction, and focused drain/merge tests.
- **Conflicting terminal fallback deletion:** Fixed at the storage layer; value equality is restored and the drain preserves a conflicting fallback instead of deleting or pointlessly folding it.
- **Repeated frozen-test nudges:** Fixed in code by tracking escalated dispatch generations.

## Security
No blocking security findings. The rework adds no attacker-controlled execution, authorization bypass, secret exposure, unsafe deserialization, or destructive operation. Gate values and lock-owner descriptions are treated as data.

## Performance
No blocking performance regression beyond the unbounded lifetime of the warning-dedupe Set noted above. Backoff remains bounded and asynchronous, delay tests use fake timers, and the patrol performs asynchronous fallback reads with record writes only for present fallbacks.

## Requirements Coverage
FR-1, FR-2 storage preservation, FR-3, FR-5, FR-6, FR-7, FR-8, and NFR-1 through NFR-4 are implemented. FR-4/NFR-5 remain incomplete because the new permanent verdict-conflict state is misclassified and activity-warning dedupe does not survive a server restart.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3092/.pan/review/agent-pan-3092-review-981f88bc/review.md

## Required action

`pan review request PAN-3092 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-3092

# Review — PAN-3092 — 2026-07-28

## Verdict: CHANGES REQUESTED — the conflict needs-you instructions cannot reliably resolve the preserved conflict, and warning dedupe still fails across a restart when the trip write is contended

## Context
- Branch: `feature/pan-3092`
- HEAD reviewed: `51f15f597ae7b2b54c2e21b4f81ce496648728c9`
- Prior reviewed HEAD: `981f88bcf04b49efad757967577a6955c3de2f65`
- Cycle scope: rework commit `51f15f597ae7b2b54c2e21b4f81ce496648728c9` only; findings below anchor to code changed since the prior verdict
- Manifest: `.pan/review/agent-pan-3092-review-51f15f59/context.json`
- Dimensions reviewed: correctness/logic, security, requirements/acceptance criteria, performance

## Blocking Findings

### ! [correctness] The conflict message’s suggested command cannot resolve several supported conflicts — `src/lib/cloister/deacon-verdict-fallback-sweep.ts:54`
**Evidence tier:** Tier 1 — static
**Changed code:** `conflictMessage()` selects only the first conflicting gate, maps it to `pan admin specialists done <specialist> ... --status <the correct value>`, and presents that as the way to apply the chosen verdict.
**Problem:** This is not a valid general conflict resolver:
- `mergeStatus` values are `merged|failed`, but `specialists done merge` accepts `passed|failed`; substituting the displayed record value `merged` is rejected.
- `inspect` requires `--item`, which the generated command omits and the pipeline-level conflict does not identify.
- `verificationStatus` has no `GATE_SPECIALIST` entry or `specialists done` specialist, and `reviewStatus: skipped` is not accepted by the review command.
- If the operator chooses the value already in the journal, re-signalling that same value normally lands no replacement fallback, so the conflicting fallback survives and every later drain withholds it again.
- A fallback can conflict on multiple gates, but the command applies only the specialist for `conflicts[0]`; the remaining conflicts keep the same generation permanently unresolved while the durable trip suppresses a fresh explanation.
**Runtime impact:** The new needs-you surface correctly says the state is permanent, then gives an instruction that can error or leave the permanent state unchanged. Operators can follow the prescribed remediation and remain stuck with the same conflict indefinitely.
**Fix:** Add a canonical conflict-resolution operation through the verdict write door that accepts the exact fallback generation and explicit per-gate choices, persists those choices, and consumes only that generation atomically. Point the needs-you message at that operation. Until such an operation exists, the only universally correct instruction is to start a strictly newer review cycle; do not claim the generic specialist command resolves every gate/source choice. Add tests for merge, inspect, verification, choosing the existing journal value, and multiple simultaneous conflicts.

### ! [requirements] A contended trip write followed by a dashboard restart still repeats the same activity warning — `src/lib/cloister/deacon-verdict-fallback-sweep.ts:145`
**Evidence tier:** Tier 1 — static
**Requirement:** FR-4 requires one activity warning per fallback episode; NFR-5 forbids a nag loop across the episode.
**Changed code:** The sweep consults `findRecoveryTrip()` before warning and uses `warnedEpisodes` only in memory. This dedupes across restart only after `recordRecoveryFailure()` successfully writes the durable trip.
**Problem:** During the exact contention state being reported, `recordRecoveryFailure()` can fail because it takes the contended record lock. The warning then exists only in the process-local Set. If the dashboard restarts before a later patrol lands the trip, the Set is lost, `findRecoveryTrip()` correctly finds no trip, and the same episode warns again. The restart test seeds a successfully written durable trip, so it does not cover the failure mode named in the previous verdict.
**Runtime impact:** Sustained lock contention plus dashboard restarts still produces repeated operator warnings for one episode, so the cross-restart no-nag requirement remains incomplete precisely when the record lock cannot provide dedupe.
**Fix:** Persist an idempotency key for the activity warning through a lock-independent durable surface, such as an idempotent activity-log write keyed by recovery path + issue + fallback generation. Add a test sequence: warning emits, trip write rejects, module state resets, the same generation is swept again, and no second activity entry is emitted.

### ! [performance] Every already-open trip still performs a no-op record write on every patrol — `src/lib/cloister/deacon-verdict-fallback-sweep.ts:165`
**Evidence tier:** Tier 1 — static
**Runtime path:** Long-lived Deacon patrol, once per patrol cycle for every unresolved fallback.
**Changed code:** The new `openTrip` read determines that an episode is already durably surfaced, but the sweep still calls `recordRecoveryFailure()` unconditionally. An open trip makes its mutator return the unchanged record (`src/lib/cloister/recovery-trip.ts:49-51`), yet `updateIssueRecord()` still writes the record and enters the commit/flush path under the per-issue lock.
**Problem:** Permanent verdict conflicts are intentionally retained until operator action, so this creates an indefinite record-lock acquisition plus filesystem/git write attempt every patrol even though the durable state is already known and unchanged. That write pressure is on the same lock family this issue is trying to stop contending.
**Runtime impact:** One unattended conflict can produce roughly one no-op record write per minute indefinitely; multiple conflicts multiply state-worktree work and can manufacture fresh lock contention. The in-memory cleanup is also inverted: a first successful trip returns `emitNeedsYou: true`, so line 171 does not delete the Set entry; if the fallback drains before another patrol reaches this block, that key remains for the server lifetime.
**Fix:** If `openTrip?.open === true`, skip `recordRecoveryFailure()` entirely and remove any matching in-process key. After any successful `recordRecoveryFailure()` call, delete the key regardless of `emitNeedsYou`; that flag controls notification, not whether the trip landed. Add assertions that the second/third patrol make zero additional record writes and that a first successful trip clears the process-local entry.

## Previous Findings Rechecked
- **Conflict classification:** Fixed. Conflicts and contention now use distinct recovery paths and messages, and the message names the gate and both values.
- **Restart dedupe after a landed trip:** Fixed. An existing durable trip suppresses a post-restart warning.
- **Frozen-agent timing test:** Fixed. The test now advances beyond the 30-minute nudge window.

## Security
No blocking security findings. Conflict values are constrained terminal-status strings, no generated command is executed automatically, and the new record reads stay behind the canonical record resolver.

## Performance
The no-op record-write loop above is blocking because it is unbounded on a long-lived patrol and adds pressure to the lock under repair. The other changed paths use asynchronous reads, bounded lists, and no synchronous shell execution.

## Requirements Coverage
The prior storage/durability findings remain fixed, and the rework distinguishes conflict from contention. FR-4/NFR-5 remain incomplete for a warning whose trip write never lands, while the new conflict surface is not yet reliably actionable across the gate set it reports.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3092/.pan/review/agent-pan-3092-review-51f15f59/review.md

## Required action

`pan review request PAN-3092 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-3092

# Review — PAN-3092 — 2026-07-28

## Verdict: CHANGES REQUESTED — the new episode ID does not make the activity warning idempotent, and the fire-and-forget write can also lose the only warning

## Context
- Branch: `feature/pan-3092`
- HEAD reviewed: `1bec60c96e1fccf50d3343a2f79799958e70c164`
- Prior reviewed HEAD: `51f15f597ae7b2b54c2e21b4f81ce496648728c9`
- Cycle scope: rework commit `1bec60c96e1fccf50d3343a2f79799958e70c164` only; the blocking finding anchors to code changed since the prior verdict
- Manifest: `.pan/review/agent-pan-3092-review-1bec60c9/context.json`
- Dimensions reviewed: correctness/logic, security, requirements/acceptance criteria, performance

## Blocking Finding

### ! [requirements/correctness] Reusing an activity ID still emits a second warning event after restart, while a failed fire-and-forget append can suppress the warning indefinitely — `src/lib/cloister/deacon-verdict-fallback-sweep.ts:170`
**Evidence tier:** Tier 1 — static

**Requirement:** FR-4 requires one activity-log warning per fallback episode with no nag loop.

**Changed code:** The sweep now derives `id: verdict-fallback:${episode}` and calls `emitActivityEntrySync()` after adding the episode to `warnedEpisodes` (`deacon-verdict-fallback-sweep.ts:170-183`). The new restart test explicitly accepts a second call after restart and checks only that both calls reuse the same ID (`tests/unit/lib/cloister/deacon-verdict-fallback-sweep.test.ts:305-334`).

**Problem:** The ID is projection identity, not append idempotency. `emitActivityEntrySync()` calls `emitActivityEntryDurable()` fire-and-forget, and that function appends a fresh `activity.entry` domain event with a fresh timestamp every time (`src/lib/activity-logger.ts:111-142`). Reusing the payload ID makes the live/bootstrap reducers replace the visible row, but it does not prevent the second event: connected clients receive it, the row gets a new timestamp and moves to the front, and the durable event log contains another warning. That is still a repeat warning for the same episode, exactly what FR-4 forbids.

The same ordering also permits loss. `warnedEpisodes` is populated before the asynchronous append has succeeded, while `emitActivityEntrySync()` swallows persistence failures. If the activity append fails and the contended recovery-trip write also fails, later patrols in the same server process suppress every retry even though neither operator surface landed. The documentation's claim that the operator-visible warning “always lands” is therefore false (`docs/AGENT-STATE-PLANES.md:93-104`).

**Runtime impact:** A dashboard restart during sustained contention re-surfaces the warning as a new activity event, while a transient event-store failure can leave the operator with no warning until another dashboard restart. The implementation provides neither at-most-once nor reliable at-least-once delivery for FR-4.

**Fix:** Make the activity write door idempotent by the episode ID so a repeated durable call does not append or publish another event, and await the durable result before treating the episode as warned. Add two integration-level assertions against the activity store/publisher rather than the mocked function: after a module restart the event count and subscriber update count remain one, and a failed first append is retried on the next patrol instead of being suppressed by `warnedEpisodes`.

## Previous Findings Rechecked
- **Conflict remediation:** The generic invalid specialist command is gone. The message names every conflicting gate and both values, explains that waiting or re-signalling cannot clear the state, and points to a strictly newer review cycle.
- **Per-patrol no-op record write:** Fixed. An already-open durable trip returns before `recordRecoveryFailure()`, while fallback draining still runs each patrol.
- **Process-local key cleanup:** Fixed. Any successful trip write removes the episode key regardless of `emitNeedsYou`.

## Advisory Findings — non-gating under the cycle-4 convergence rule
- `src/cli/commands/specialists/done.ts:354` infers the current signal's outcome from any issue-level fallback, so an unrelated older gate can produce incorrect “do not re-run” guidance. This code predates the current rework commit and is not promotable in this cycle.
- `src/lib/cloister/deacon.ts:1483` lets a rejected frozen-test escalation abort processing of the remaining statuses for that patrol. This code predates the current rework commit and is not promotable in this cycle.
- `src/lib/cloister/test-agent-queue.ts:56` uses `curl -s`, which exits zero on HTTP 4xx/5xx and can cause the generated test prompt to report the wrong single-attempt outcome. This code predates the current rework commit and is not promotable in this cycle.
- The conflict message should name the concrete existing recovery command, `pan review restart <issueId>`, so “dispatch a strictly newer review cycle” is fully actionable from the message alone. The current wording is truthful, so this is advisory rather than a shipping blocker.

## Security
No blocking security findings. Conflict values remain constrained terminal-status strings, no operator text is executed, and the rework adds no lock bypass, cross-machine verdict transport, or direct canonical-state mutation.

## Performance
The prior unbounded no-op write loop is fixed. Repeated same-ID activity appends are low-volume, but they are still unnecessary durable events and live publications; the blocking impact is the violated exactly-once operator contract rather than throughput.

## Requirements Coverage
The rework fixes the prior conflict-remediation and record-write-loop blockers. FR-4 remains incomplete because the activity warning is appended again after restart and may be silently lost before restart; NFR-1 through NFR-5 and the other functional requirements have no new blocking regression in this cycle.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3092/.pan/review/agent-pan-3092-review-1bec60c9/review.md

## Required action

`pan review request PAN-3092 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-3092

# Review — PAN-3092 — 2026-07-28

## Verdict: CHANGES REQUESTED — the new idempotent activity API is bypassed by the production Deacon transport and cannot detect real event-store write failures

## Context
- Branch: `feature/pan-3092`
- HEAD reviewed: `4db74b5910033b9d19d1f4b3f99dea7fbf1a2225`
- Prior reviewed HEAD: `1bec60c96e1fccf50d3343a2f79799958e70c164`
- Cycle scope: rework commit `4db74b5910033b9d19d1f4b3f99dea7fbf1a2225` only; the blocker below was introduced by this rework and could not have been raised in cycle 1
- Manifest: `.pan/review/agent-pan-3092-review-4db74b59/context.json`
- Dimensions reviewed: correctness/logic, security, requirements/acceptance criteria, and performance

## P1 (Major) — Blocking

### The production sweep still appends repeat warnings after restart and marks failed writes as durable — `src/lib/activity-logger.ts:163`

**Evidence tier:** Tier 1 — static, traced through the production provider wiring.

**Requirement:** FR-4 requires one activity-log warning per fallback episode with no nag loop. The current rework claims `emitActivityEntryOnce()` provides that guarantee and reports `appended | duplicate | failed` only after durability.

**Changed code:** `emitActivityEntryOnce()` makes `hasEventWithPayloadId` optional, skips the duplicate check when it is absent, awaits `appendAsync()`, and then returns `appended` (`src/lib/activity-logger.ts:163-177`). The sweep treats that outcome as durable and adds the episode to `warnedEpisodes` (`src/lib/cloister/deacon-verdict-fallback-sweep.ts:170-188`).

**Problem:** The production caller is the Deacon child, and its provider is explicitly the HTTP-backed `createDeaconEventClient()` (`src/dashboard/server/deacon-main.ts:55-73`). That client has no `hasEventWithPayloadId` method (`src/lib/cloister/deacon-event-client.ts:16-22`), while its `appendAsync()` only enqueues locally and immediately resolves `0` before any HTTP request or SQLite commit (`src/lib/cloister/deacon-event-client.ts:152-160`). Consequently:

1. After a dashboard/Deacon restart, the process-local set is empty, the production provider cannot query the prior ID, and `emitActivityEntryOnce()` appends the same warning again. The documented “degrades to at-least-once” path is the actual FR-4 path, so the implementation deliberately permits the repeat event the previous verdict blocked.
2. A queued event can later fail HTTP delivery, remain buffered, be dropped on buffer overflow, or disappear with the child process, but `emitActivityEntryOnce()` has already returned `appended`; the sweep marks the episode warned and suppresses its own retry.
3. Even the in-process SQLite provider cannot reliably return `failed`: the async batch catch resolves every queued promise with sequence `0` rather than rejecting (`src/dashboard/server/event-store.ts:237-257`), so `emitActivityEntryOnce()` returns `appended` after a failed transaction. The new unit test uses a fake `appendAsync` that throws, which does not model the real store contract (`src/lib/__tests__/activity-logger.test.ts:105-162`).
4. The direct-store `hasEventWithPayloadId()` check and later `appendAsync()` are separate operations, so concurrent callers can both observe absence and append. There is no unique constraint or atomic insert enforcing the stated at-most-once contract (`src/dashboard/server/event-store.ts:204-208,367-369`).

**Runtime impact:** The exact production patrol remains capable of publishing multiple activity warnings for one fallback episode, while a failed activity write can still be treated as durable and never retried. The new code therefore does not satisfy FR-4 and its public `ActivityEmitOutcome` can report the opposite of the real persistence result.

**Fix:** Put the idempotency decision and append behind one atomic event-store write operation, then expose that operation through the Deacon HTTP client. The server must commit-or-detect-duplicate by the episode key in one transaction and return `appended`, `duplicate`, or `failed` only after that transaction settles; the child must await the server response rather than treating local enqueue as durability. Add integration tests using the real event store plus the Deacon client transport for: restart dedupe, transaction failure, HTTP failure/retry, and two concurrent calls for the same ID. The sweep should add `warnedEpisodes` only on a server-confirmed appended/duplicate result.

## Previous Finding Rechecked
- The rework now awaits a result before updating the process-local set and includes retry behavior for a mocked failed append. That is the right caller-side shape, but the production event transport does not implement the result contract, so the prior FR-4 blocker remains at the write-door boundary.
- Conflict remediation, open-trip no-op suppression, UAT durability, content-aware fallback preservation, and terminal-verdict backoff have no new regression in this rework commit.

## Security
No blocking security finding. The new SQL uses bound parameters, the activity ID is data rather than executable input, and the rework adds no authorization bypass, direct canonical-state write, lock exemption, or cross-machine verdict transport.

## Performance
`hasEventWithPayloadId()` scans all retained events of one type and evaluates `json_extract` because the payload ID is not indexed; this is acceptable at the low warning frequency but would not scale as a generic hot-path idempotency primitive. The atomic server-side fix should use an indexed uniqueness key rather than a check-then-scan query.

## Requirements and Acceptance Criteria
The latest rework does not meet FR-4's one-warning/no-nag guarantee on the actual Deacon path. The added tests validate enhanced fakes rather than the production HTTP client and event-store failure semantics, so they cannot mechanically prove the requirement. No new violation was found for NFR-1 through NFR-5 or the other functional requirements in the six-file rework delta.

## Advisory Findings
The previously supplied CodeRabbit findings target code unchanged since the prior review HEAD and remain non-gating under the cycle-2+ convergence rule. They should be handled as follow-up work rather than promoted during this rework cycle.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3092/.pan/review/agent-pan-3092-review-4db74b59/review.md

## Required action

`pan review request PAN-3092 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-3092

# Review — PAN-3092 — 2026-07-28

## Verdict: CHANGES REQUESTED — the settled append request has no deadline and can wedge the entire Deacon patrol while its heartbeat continues reporting healthy

## Context
- Branch: `feature/pan-3092`
- HEAD reviewed: `f110307644701e83f5413011b79bea06fb74cd73`
- Prior reviewed HEAD: `4db74b5910033b9d19d1f4b3f99dea7fbf1a2225`
- Cycle scope: rework commit `f110307644701e83f5413011b79bea06fb74cd73` only; the blocker is in the new HTTP settlement path and could not have been raised before this commit introduced that path
- Manifest: `.pan/review/agent-pan-3092-review-f1103076/context.json`
- Dimensions reviewed: correctness/logic, security, requirements/acceptance criteria, and performance

## P1 (Major) — Blocking

### A hung `/append-once` request wedges all later patrol work without tripping the Deacon health watchdog — `src/lib/cloister/deacon-event-client.ts:169`

**Evidence tier:** Tier 1 — static, traced through the production patrol scheduler and heartbeat.

**Changed code:** `DeaconEventClient.appendOnce()` performs a direct `fetch()` and awaits both the response and `response.json()`, but supplies no `AbortSignal` or other deadline (`src/lib/cloister/deacon-event-client.ts:169-192`). `sweepStrandedVerdictFallbacks()` awaits that call inline while iterating issues (`src/lib/cloister/deacon-verdict-fallback-sweep.ts:176-191`).

**Problem:** A connection that is accepted but does not complete can hold this promise until the runtime's long transport timeout, with no application-level bound. This stalls `runPatrol()` at the first old fallback and prevents every remaining issue and every patrol phase after the sweep from running. Scheduled patrols are then skipped because the prior patrol remains in flight (`src/lib/cloister/deacon.ts:3229-3246`). At the same time, `runPatrol()` keeps `startPatrolHeartbeatTicker()` running until its `finally` block, so the stalled patrol refreshes its heartbeat every half interval and the supervisor sees it as healthy (`src/lib/cloister/deacon.ts:414-416,2616,3220-3222`). The new recovery path can therefore disable Deacon indefinitely while defeating the watchdog intended to detect exactly that condition.

The test named “reports failed when the request never completes” does not exercise this state: its fetch mock throws `ECONNREFUSED` immediately (`tests/unit/lib/cloister/deacon-event-client.test.ts:201-205`). A never-settling promise would leave that test and the production patrol pending forever.

**Runtime impact:** During the degraded state this feature is meant to surface, an overloaded, restarting, or event-loop-starved dashboard can make the Deacon child wait on the new loopback request. Agent recovery, fallback draining for later issues, lifecycle cleanup, and subsequent patrols stop, while the heartbeat continues to claim the Deacon is alive. Manual patrol IPC can also start additional unguarded `runPatrol()` calls in the child, compounding the blocked work.

**Fix:** Give `appendOnce()` a bounded request deadline shorter than the patrol interval, abort the fetch/body read on expiry, and return `failed` so the existing sweep logic retries next patrol. A timeout after a server commit is safe: retrying the same idempotency key returns `duplicate`. Add a fake-timer test whose fetch promise truly never settles until its signal aborts, and assert the client returns `failed` at the configured deadline with nothing added to the batch queue. The timeout must cover both response headers and body parsing.

## Previous Finding Rechecked
- Fixed: both production activity providers now expose a settled `appendOnce` path.
- Fixed: the SQLite store claims the idempotency key and inserts the event in one transaction, so concurrent retries cannot append twice.
- Fixed: the Deacon child uses a dedicated HTTP endpoint and does not treat local queueing as durability.
- Fixed: transaction failure throws, the HTTP client reports `failed`, and the sweep leaves the episode eligible for retry.

## P2 (Minor) — Non-blocking

- `emitActivityEntryOnce()` documents `unconfirmed` as an outcome callers requiring delivery must not treat as success, and `docs/AGENT-STATE-PLANES.md:121-125` says the sweep marks only `appended`/`duplicate` as warned. The sweep currently treats every outcome except `failed` as success, so `unconfirmed` is added to `warnedEpisodes` at `src/lib/cloister/deacon-verdict-fallback-sweep.ts:184-190`. Both production providers implement `appendOnce`, so this does not block the production path, but the code and contract should agree: handle `unconfirmed` like `failed` or remove the fallback outcome.
- `event_idempotency` is never compacted when seven-day event retention deletes the corresponding event. Warning frequency is low, so this is not an immediate performance blocker, but the table is monotonically growing and its orphaned keys permanently suppress recreation after the activity event itself is compacted. Add lifecycle cleanup tied to event retention or document the intentionally permanent key retention.

## Security
No blocking security findings. The new endpoint reuses constant-time internal-token validation and loopback-origin enforcement, validates the event shape and non-empty key, uses bound SQL parameters, and exposes no unauthenticated state mutation. Database error details are returned only across the authenticated internal route.

## Performance
The atomic insert is low-frequency and indexed by a primary key, avoiding the prior JSON scan. The missing network deadline is the blocking performance/availability issue because it stalls the serial patrol while the heartbeat masks the stall. Synchronous SQLite work in the route is small and transactionally bounded under normal operation.

## Requirements and Acceptance Criteria
The new transaction and HTTP round trip close the prior at-most-once and false-durability gaps for FR-4. FR-5 remains incomplete because the new awaited request can prevent the sweep from reaching later issues or returning for future patrol cycles. No new violation was found for the lock-protocol constraint, verdict fabrication rules, cross-machine non-goal, or the remaining functional requirements.

## Advisory Context
The supplied CodeRabbit findings concern code unchanged since the prior review HEAD and remain non-gating under the cycle-2+ convergence rule. They should be handled as follow-up work rather than promoted during this rework cycle.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3092/.pan/review/agent-pan-3092-review-f1103076/review.md

## Required action

`pan review request PAN-3092 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-3092

# Review — PAN-3092 — 2026-07-28

## Verdict: CHANGES REQUESTED — the idempotency table is created in the wrong database schema, so the production append-once endpoint cannot append anything

## Context
- Branch: `feature/pan-3092`
- HEAD reviewed: `6986a22ca9666f15c36f1b58d62cfbd56043be38`
- Prior reviewed HEAD: `f110307644701e83f5413011b79bea06fb74cd73`
- Cycle scope: current rework plus the production database-opening path required to validate its new compaction and timeout behavior
- Manifest: `.pan/review/agent-pan-3092-review-6986a22c/context.json`
- Dimensions reviewed: correctness/logic, security, requirements/acceptance criteria, and performance

## P0 (Critical) — Blocking

### `event_idempotency` is added to `panopticon.db`'s schema, but `appendOnce()` runs against `overdeck.db`, where the table is never created — `src/lib/database/schema.ts:429`

**Evidence tier:** Tier 1 — static, traced through both database openers and the production event-store initialization.

**Changed behavior:** `EventStore.appendOnce()` prepares statements against `event_idempotency`, and the current rework also calls that statement family during startup compaction (`src/dashboard/server/event-store.ts:219-241,369-382,397-428`). The branch adds the table only inside `src/lib/database/schema.ts`.

**Problem:** Those are different databases and migration systems:

- `src/lib/database/schema.ts` is used by `src/lib/database/index.ts` for `~/.overdeck/panopticon.db` (`src/lib/database/index.ts:33-75`).
- The dashboard event store imports `getOverdeckDatabasePath()` and `getOverdeckDatabaseSync()`, which open `~/.overdeck/overdeck.db` (`src/lib/overdeck/paths.ts:5-10`, `src/dashboard/server/event-store.ts:144-153`).
- `overdeck.db` is initialized from `drizzle/overdeck/0000_overdeck_init.sql` only when the `agents` table is absent; existing databases return without rerunning that file, then receive explicit runtime top-ups (`src/lib/overdeck/infra.ts:103-154,292-314`). Neither the init SQL nor a runtime top-up creates `event_idempotency`.

A repository-wide search finds no production `CREATE TABLE event_idempotency` outside the unrelated `panopticon.db` schema. The event-store tests hide this by manually creating the table in `makeDb()` (`tests/dashboard/event-store.test.ts:16-34`) instead of opening a production-shaped `overdeck.db`.

**Runtime impact:** On every real installation, including a fresh one, the first `/api/internal/events/append-once` call fails while preparing the claim statement and returns HTTP 500. `emitActivityEntryOnce()` therefore returns `failed` on every patrol. The one activity warning required by FR-4 never lands; the new at-most-once transport is wholly unreachable. Startup compaction silently catches the same missing-table error, so boot appears healthy and gives no diagnostic before the feature fails.

**Fix:** Create the table through the `overdeck.db` schema door: add it to `drizzle/overdeck/0000_overdeck_init.sql` for fresh databases and add an idempotent runtime top-up in `src/lib/overdeck/infra.ts` for existing databases, including the schema-audit expectation. Remove the unrelated table from `src/lib/database/schema.ts` unless `panopticon.db` genuinely needs it. Add a migration test that starts from an existing production-shaped `overdeck.db` with `agents` present, runs `getOverdeckDatabaseSync()`, and proves `appendOnce()` succeeds; a hand-built test database that pre-creates the table does not test the shipping path.

This finding is independently confirmed from the code paths above; it is not promoted merely because CodeRabbit mentioned it. The table contract did not exist in the original cycles, and the current compaction rework made tracing its lifecycle through the production opener necessary.

## P1 (Major) — Blocking

### A 10-second deadline per issue does not bound the serial sweep to one patrol interval — `src/lib/cloister/deacon-event-client.ts:30`

**Evidence tier:** Tier 1 — static.

The new request timeout correctly bounds one `appendOnce()` call, but `sweepStrandedVerdictFallbacks()` iterates all in-pipeline issues sequentially and awaits each warning before advancing. If the dashboard accepts connections but stalls them, six old fallbacks consume the full 60-second patrol interval; dozens can hold the patrol for many minutes. Scheduled patrols are skipped while the previous one remains in flight, and the heartbeat ticker continues reporting healthy throughout. The exact systemic dashboard degradation likely to strand several verdicts is also the state likely to make every loopback request hit its deadline.

**Fix:** Bound the warning work at the sweep/patrol level rather than multiplying a per-request budget by the number of issues. Use a shared deadline or bounded concurrent fan-out so all pending warning attempts settle within a fixed budget below the patrol interval, while preserving the idempotency key and treating timed-out calls as retryable `failed` outcomes. Add a fake-timer test with at least six simultaneously hanging warning requests and assert the whole sweep returns within the global budget and later patrol phases remain reachable.

## Previous Finding Rechecked
- Fixed for a single request: one `AbortController` now covers response headers and body parsing, and timeout returns `failed` without queueing the event.
- Fixed: `unconfirmed` is treated like `failed`, so the process-local set does not suppress an unverified warning.
- Fixed: idempotency keys are compacted with the seven-day activity retention, assuming the table exists in the correct database.
- The atomic claim/insert and settled HTTP response remain logically sound once the production schema makes them reachable.

## Security
No blocking security findings. The internal endpoint remains token-authenticated with constant-time comparison and loopback-origin enforcement; SQL values are bound parameters; retries cannot duplicate an event after a committed-but-unacknowledged request. The missing table is an availability/data-integrity failure rather than an authorization flaw.

## Performance
The primary performance blocker is the multiplicative timeout in the serial sweep. The primary-key claim avoids the prior JSON scan, and seven-day claim compaction bounds table growth once the table is migrated into `overdeck.db`.

## Requirements and Acceptance Criteria
FR-4 is not met because the production event store lacks the table required to emit the warning. FR-5 is not fully met because N sequential 10-second warning calls can consume N×10 seconds and prevent a patrol from sweeping the remaining issues or reaching later phases. The current rework otherwise preserves the lock protocol, avoids verdict fabrication, uses fake timers for timeout tests, and introduces no cross-machine verdict transport.

## Advisory Context
The older CodeRabbit findings target code unchanged since the prior review and remain non-gating under the convergence rule. The database finding above is blocking because direct inspection confirms a production P0 and the current event-store tests/compaction changes exercise the same newly introduced table contract without testing its actual schema door.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3092/.pan/review/agent-pan-3092-review-6986a22c/review.md

## Required action

`pan review request PAN-3092 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-3092

# Review — PAN-3092 — 2026-07-28

## Verdict: CHANGES REQUESTED — the new shared warning budget permanently starves later issues when an earlier episode keeps timing out

## Context
- Branch: `feature/pan-3092`
- HEAD reviewed: `767b20a2125267b798422a20769185b418eff205`
- Prior reviewed HEAD: `6986a22ca9666f15c36f1b58d62cfbd56043be38`
- Cycle scope: rework commit `767b20a2125267b798422a20769185b418eff205` only; the blocker is introduced by the new shared-budget algorithm and could not have been raised in cycle 1
- Manifest: `.pan/review/agent-pan-3092-review-767b20a2/context.json`
- Dimensions reviewed: correctness/logic, security, requirements/acceptance criteria, and performance

## P1 (Major) — Blocking

### The shared budget always favors the first old fallback, so later episodes may never receive their required warning — `src/lib/cloister/deacon-verdict-fallback-sweep.ts:133`

**Evidence tier:** Tier 1 — static and locked by the new test expectation.

**Requirement:** FR-4 requires one activity-log warning per fallback episode, and FR-5 requires the patrol to sweep every pending issue each cycle.

**Changed code:** The sweep sets one 30-second deadline, processes `Object.entries(statuses)` serially, and starts a warning only while budget remains (`deacon-verdict-fallback-sweep.ts:133-145,216-239`). Once the first hanging attempt consumes the budget, every later issue receives a synthetic `failed` result without calling `emitActivityEntryOnce()` at all.

**Problem:** Iteration order is stable across patrols. If the first episode's warning request keeps hanging and its recovery-trip write also keeps losing the contended record lock — the exact hard case this feature is designed for — that same first episode consumes the whole budget every patrol. Nothing marks it warned or opens a durable trip, so later episodes remain behind it forever and never even attempt the activity warning. Returning an action string saying their warning “could not be confirmed” is not an operator-visible activity warning.

The new test explicitly codifies the starvation: seven issues are eligible, but it expects `emitActivityEntryOnce` to be called exactly once (`tests/unit/lib/cloister/deacon-verdict-fallback-sweep.test.ts:562-570`). Its preceding assertion that every issue was “visited” only proves each drain ran; it does not prove FR-4 warning delivery was attempted for each episode.

**Runtime impact:** A systemic dashboard stall plus record-lock contention on the first listed issue can suppress warnings for every later stranded verdict indefinitely. The sweep now returns within budget, but it does so by silently dropping the required warning work in a deterministic, unfair order.

**Fix:** Bound latency without dropping later episodes. Collect all eligible warning candidates first, then settle them through a bounded operation that gives every candidate an attempt within the shared deadline — for example, a single authenticated batch append-once request or concurrent idempotent requests under one abort deadline. If concurrency must be capped, carry a fair rotating cursor across patrols so an unresolved first episode cannot monopolize every cycle, and state explicitly when work is deferred. Add a multi-patrol fake-timer test where the first episode's warning and trip both fail repeatedly, and prove every later issue still reaches `emitActivityEntryOnce()` and eventually appends exactly one warning.

## Previous Findings Rechecked
- **Production schema door:** Fixed. Fresh `overdeck.db` files create `event_idempotency` through `drizzle/overdeck/0000_overdeck_init.sql`; existing databases receive the idempotent top-up in `src/lib/overdeck/infra.ts`; the unrelated `panopticon.db` table was removed; and production-opener tests cover fresh, upgraded, and append-once paths.
- **Single-request wedge:** Fixed. The Deacon HTTP client applies a 10-second abort deadline to headers and body parsing and returns `failed` without queueing.
- **Unconfirmed outcome:** Fixed. `unconfirmed` remains eligible for retry rather than entering the process-local warned set.
- **Claim retention:** Fixed. Idempotency claims age out with seven-day event retention.

## P2 (Minor) — Non-blocking

- `SWEEP_WARNING_BUDGET_MS` is hardcoded to 30 seconds while `patrolIntervalMs` is configurable and not clamped to 60 seconds. A configured interval below 30 seconds reopens overlapping/skipped patrol behavior. Derive the budget from the resolved patrol interval or enforce a minimum interval if the 60-second assumption is contractual.
- The new `event_idempotency` table is not listed in `OVERDECK_SCHEMA_TOP_UP_EXPECTATIONS`, so the schema audit cannot report its absence if the top-up fails. `runSchemaTopUp` logs the immediate failure, but adding the table expectation would preserve the existing drift-audit contract.

## Security
No blocking security findings. The migration adds no external surface; the append-once route remains internal-token authenticated and loopback-restricted; keys and event values use bound SQL parameters; retries remain idempotent and cannot fabricate verdicts.

## Performance
The production schema fix and indexed primary-key claim are efficient. The shared budget prevents unbounded serial latency, but the current first-item-wins policy trades the wedge for deterministic starvation. A batch append-once request would bound wall-clock cost and avoid an N-request loopback burst while preserving one attempt per eligible episode.

## Requirements and Acceptance Criteria
FR-4 and FR-5 remain incomplete because later eligible episodes can be skipped every patrol without an activity-warning attempt. The schema correction now makes the append-once path reachable on fresh and existing installations, and the rework preserves NFR-1 through NFR-5, the no-fabrication constraint, and all other functional requirements.

## Advisory Context
The older CodeRabbit findings concern code unchanged since the previous review and remain non-gating under the convergence rule. The missing-schema finding is resolved by this commit and was reverified through the actual `overdeck.db` opener rather than the hand-built event-store fixture.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3092/.pan/review/agent-pan-3092-review-767b20a2/review.md

## Required action

`pan review request PAN-3092 -m "Fixed review issues"`

Issue #1990, opened 2026-06-20
First-class workspaces and projects with per-workspace memory


eltmon:
**PRD:** [`drafts/pan-1990.md`](https://github.com/eltmon/overdeck/blob/overdeck-state/drafts/pan-1990.md) — the PRD is the implementation source of truth; it carries verified file/line references, decisions, numbered work items, and supersedes this issue body where they differ (notably: per-issue memory already exists and is re-keyed, not rebuilt; no separate `session-context.md` artifact; search scopes are flags on `pan memory search`; `destroy --purge-memory` instead of `delete`).

## Summary

Promote workspaces and projects to first-class entities in Overdeck. Today "workspace" is implicit — a `workspaces/feature-<issue-id>/` directory pinned 1:1 to a GitHub issue. This change introduces an explicit **project** (persistent identity for the thing being worked on) and **workspace** (one in-flight session against a project), with a `kind` discriminator so non-issue session units can coexist with the existing issue pipeline. Every workspace gets its own **memory home** — `session-context.md` + a rolling observation log + FTS search — so agents can recall what was discussed in earlier sessions instead of starting cold.

This is delivered as **one complete feature** — schema, migration, memory pipeline, CLI, and dashboard surface together. Sequencing of the work is left to the implementor; there are no prescribed phases and no partial-delivery milestones.

## Motivation

The current model has three sharp limits:

1. **You can't use Overdeck without an issue.** Exploring an idea, doing scratch research, or just working on `main` bypasses every Overdeck affordance (no dashboard surface, no spawn flow, no review pipeline). Filing an issue first is friction; not filing one loses every Overdeck capability.
2. **Memory is pinned to issues.** A per-issue memory subsystem now exists (`src/lib/memory/` — observations, FTS, prompt-time injection, rollup), keyed `~/.overdeck/memory/{projectId}/{issueId}/`. But because the key is the issue, non-issue sessions (working on `main`, scratch exploration, plain conversations) capture nothing and recall nothing. Re-keying memory to the workspace — with issue workspaces as one kind — gives every session unit rolling, searchable recall.
3. **No layout persistence on a session unit.** Dashboard tabs/panes/conversations are ephemeral; closing a session loses the workspace's editor state.

## Mental model

- **Project** = the long-lived noun. A stable id, a primary path on disk, optional secondary target paths, pinned reference docs. Outlives every individual attempt at work.
[...]

PR #3277, opened 2026-07-29
PAN-1990

eltmon:
**Issue:** #1990

## Acceptance Criteria

- [ ] Schema migration: projects, workspaces, project_targets, pinned_docs tables + workspace_id columns
- [ ] Two doors: workspaces resolver and writer in src/lib/workspaces/
- [ ] Guard script: build fails on direct SQL against workspace tables outside the doors
- [ ] Boot seeding: projects.yaml upsert + issue-worktree backfill at dashboard boot
- [ ] pan admin db rebuild-workspaces command
- [ ] pan start / pan plan create the issue workspace row before worktree creation
- [ ] Close-out archives the workspace row; deep-wipe never touches memory homes
- [ ] Conversations carry workspace_id resolved from cwd
- [ ] Resource discovery and idle-stack reaper join through the workspaces resolver
- [ ] pan workspace new + pan workspace main verbs
- [ ] pan workspace get/activate/archive + list filters + destroy --purge-memory
- [ ] pan project add-target
- [ ] Contracts: MemoryIdentity/MemoryObservation issueId becomes nullable
- [ ] Memory paths re-key to workspaceId + identity record module
- [ ] Migrate issue-keyed memory homes to workspace UUIDs + FTS re-point
- [ ] Observation capture + injection for main/scratch conversations
- [ ] pinned_docs: pan memory pin/unpin CLI
- [ ] pan memory search --workspace/--project/--global flags
- [ ] Durable summaries + pins mirrored to the new memory state domain
- [ ] pan memory backfill from Claude JSONL transcripts
- [ ] Workspaces REST API route module
- [ ] Domain-event payloads additively carry workspaceId
- [ ] Sidebar Workspaces section above Projects
- [ ] Workspace view route with terminals, conversations, memory surface
- [ ] Cmd-K workspace switcher
- [ ] No-loss audit: surface inventory + gating test
- [ ] Documentation + wrapper skills


eltmon:
# Review CHANGES REQUESTED for PAN-1990

# Review — PAN-1990 — 2026-07-29

## Verdict: CHANGES REQUESTED — multiple stated acceptance criteria are incomplete, and several primary paths are broken

## Context
- Branch: `feature/pan-1990`
- HEAD reviewed: `ae1895419650db87d0d01d9f3d2c41d6f8febe8c`
- Base merge-point: `ed7d6766f3ef2a788daf4d9ead51139fc4ef1a65`
- PR: `#3277`
- Review cycle: 1
- Coverage: all four high-risk files, all five medium-risk files, the workspace resolver/writer/rebuild core, lifecycle integrations, CLI commands, memory migration/search/backfill/mirroring, event contracts, dashboard sidebar/workspace view, and their focused tests. No high- or medium-risk file was left uncovered.

## Blocking Findings

### ! [requirements/correctness] `agents.workspace_id` is explicitly omitted — `src/lib/overdeck/infra.ts:70-79`, `src/lib/overdeck/infra.ts:272-276`
AC-1 and FR-4 require both `conversations.workspace_id` and `agents.workspace_id` on fresh and migrated databases. The implementation deliberately adds only the conversation column, and the schema test explicitly documents the agent column as deferred. A fresh or upgraded database therefore fails the stated schema contract, and the new workspace view has to fall back to an issue-id guess instead of the required agent/workspace linkage.

### ! [requirements] The two-door guard allows SQL outside the two door files and misses JOIN reads — `scripts/guard-workspace-doors.sh:15-22`, `scripts/guard-workspace-doors.sh:37-43`
FR-3/AC-2 require the lint gate to reject table SQL everywhere except `resolver.ts`, `writer.ts`, and schema/top-up code. The allowlist excludes every `src/lib/workspaces/*.ts` file, so direct SQL added to `rebuild.ts`, `types.ts`, or any future sibling passes. The patterns also omit `JOIN workspaces|projects|project_targets|pinned_docs`, so a direct join outside the doors passes too. The guard test covers only `FROM workspaces` in an unrelated directory and does not lock the actual invariant.

### ! [requirements/correctness] `pan start`/`pan plan` do not guarantee a workspace row before worktree creation — `src/lib/workspace-manager/create.ts:204-225`
FR-6/AC-4 require the issue workspace row to exist before the worktree is created. This path silently skips the write when the project table has not already been boot-seeded, and it catches writer failures and continues into worktree creation. On a fresh database, a CLI-created worktree can therefore exist without a row until a later dashboard backfill, which is the exact ordering the criterion forbids.

### ! [requirements/correctness] The legacy memory-home migration is not wired into boot — `src/lib/workspaces/rebuild.ts:17-23`, `src/dashboard/server/main.ts:655-659`
WI-6 requires `migrateMemoryHomesToWorkspacesOnce()` to run at boot behind the marker. The implementation calls only project seeding and issue-worktree backfill; its own module header labels boot migration as a follow-up. After upgrade, existing issue-keyed observations and status remain under `{projectId}/{issueId}`, while all new readers use `{projectId}/{workspaceUuid}`, so users lose their existing memory until they know to run the admin rebuild command manually.

### ! [correctness/data preservation] Migration cannot merge a legacy home into an active UUID home — `src/lib/workspaces/rebuild.ts:145-160`
`mergeDirectoryEntries()` renames each top-level entry directly onto the destination. If the UUID home has already received a new `observations/`, `pending/`, `summaries/`, or `rag-runs/` directory before migration, renaming the legacy directory onto the existing directory fails with `ENOTEMPTY`. This is especially likely because boot migration is absent. The migration then aborts before FTS re-pointing, so AC-6/NFR-5's idempotent, non-destructive migration does not hold for a realistic upgraded workspace.

### ! [correctness/security] `pan workspace new --isolated` both fails on the normal branch and constructs a shell-injectable command — `src/cli/commands/workspace-scratch.ts:62-77`
The inferred parent is normally the project's currently checked-out branch. `git worktree add <path> <parentBranch>` tries to check that same branch out in a second worktree and fails with `fatal: '<branch>' is already used by worktree ...`; a local reproduction returned exit 128. The command also interpolates the user-controlled workspace name/path and `--parent-branch` into a shell string, so quotes or shell metacharacters can execute arbitrary commands. AC-5's isolated scratch-workspace path is therefore unusable and unsafe; use an argument-vector spawn and create a distinct scratch branch.

### ! [correctness/requirements] Workspace conversation filtering uses cwd containment instead of `workspace_id` — `src/dashboard/frontend/src/components/workspace/WorkspaceView.tsx:120-128`
The default “This workspace” filter includes every conversation whose cwd is beneath the workspace path. A main workspace rooted at `/repo` therefore includes conversations from `/repo/workspaces/feature-*`, even though those rows belong to separate issue workspaces. FR-15/AC-11 require workspace-scoped conversations using the new linkage; the view must filter by `conversation.workspaceId`, with cwd matching only as an explicit legacy fallback.

### ! [correctness/requirements] Query-expansion logs keep writing to the retired issue-keyed home and drop main/scratch logs — `src/lib/memory/query-expansion.ts:246-250`
`resolveRagRunsFile()` now takes a workspace UUID, but this call passes `identity.issueId` and returns early when the issue is null. After a successful migration, each issue prompt recreates `{projectId}/{issueId}/rag-runs`, splitting one workspace's memory across old and new homes; main/scratch prompts never record expansion decisions. This violates FR-7's workspace-keyed layout and undermines migration idempotency.

### ! [requirements/performance] The new search scopes do not use FTS and `--workspace` does not resolve its owning project — `src/lib/memory/cli.ts:71-90`, `src/lib/memory/cli.ts:190-207`
FR-9 requires `--workspace`, `--project`, and `--global` against the existing FTS. Instead, global search enumerates every workspace directory, reads every daily JSONL file into memory, parses all observations, sorts the full corpus, and only then keeps the requested limit. Runtime and memory therefore grow with the complete history rather than the result set. Separately, `--workspace <id>` still searches the default `overdeck` project unless the caller also supplies `--project`, so a valid workspace in any other project returns no results; workspace name resolution from D-3 is also absent.

### ! [requirements] Event schemas accept `workspaceId`, but event producers still omit it — `packages/contracts/src/events.ts:54-58`
D-10/WI-11 require issue-bearing domain-event payloads to carry the workspace id. This change makes the field optional in contracts, but representative production emitters such as `src/dashboard/server/routes/agents/lifecycle-restart.ts:118-136` still emit only `issueId`; the branch-wide producer search found population only for `operator.intervention`. Consumers therefore receive the same payloads as before, and AC-11 is not met.

### ⊗ [security] Pinned docs can escape the project root and inject arbitrary local files into model context — `src/cli/commands/memory.ts:203-210`, `src/lib/memory/injection.ts:485-509`
An absolute path outside the project is converted to a `../../...` relative path, and a relative traversal is accepted unchanged. Injection later resolves that path against the project root without a containment check and reads the resulting file. `pan memory pin /home/user/.ssh/id_rsa` can therefore persistently place a private key or any other readable local file into prompt-time context sent to the configured model provider. Pin creation must reject paths whose resolved target is outside the project's allowed documentation root.

### ! [requirements/durability] Deleting a workspace removes pin rows but leaves their committed mirrors — `src/lib/workspaces/writer.ts:157-169`
FR-10 requires workspace-scoped pins to be removed with the workspace, and FR-11 makes the state-domain mirror the durable representation. `deleteWorkspace()` deletes the SQLite rows inside a synchronous transaction but never calls `unmirrorPin`, so `memory/pins/` retains stale committed descriptors after destruction. The two stores diverge immediately, and any future recovery from the durable state resurrects pins that the workspace deletion was supposed to remove.

## Non-blocking Findings

### ~ [correctness] Malformed JSON silently becomes a destructive default — `src/dashboard/server/routes/workspace-registry.ts:57-64`, `src/dashboard/server/routes/workspace-registry.ts:137-150`
`readJsonBody` converts invalid JSON to `{}`. The archive route interprets a missing `archived` field as `true`, and the favorite route similarly defaults to `true`, so a malformed mutation body succeeds with a state change instead of returning 400.

### ~ [durability] Pin mirror filenames are collision-prone — `src/lib/memory/state-mirror.ts:83-85`
Replacing `/` with `__` makes `a/b.md` and `a__b.md` map to the same committed filename. Pinning one can overwrite the other's durable descriptor, and unpinning either can delete the mirror for both. Use the pin UUID or a collision-resistant encoding/hash in the mirror filename.

## Clean Dimensions

- No use of `execSync` was added to dashboard-server-reachable code.
- JSONL backfill reads transcript files without modifying, truncating, moving, or deleting them.
- The REST mutation routes apply the existing dashboard mutation-origin/token guard.
- `git diff --check origin/main...HEAD` passed.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-1990/.pan/review/agent-pan-1990-review-ae189541/review.md

## Required action

`pan review request PAN-1990 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1990

# Review — PAN-1990 — 2026-07-29

## Verdict: CHANGES REQUESTED — multiple stated acceptance criteria are incomplete, and several primary paths are broken

## Context
- Branch: `feature/pan-1990`
- HEAD reviewed: `ae1895419650db87d0d01d9f3d2c41d6f8febe8c`
- Base merge-point: `ed7d6766f3ef2a788daf4d9ead51139fc4ef1a65`
- PR: `#3277`
- Review cycle: 1
- Coverage: all four high-risk files, all five medium-risk files, the workspace resolver/writer/rebuild core, lifecycle integrations, CLI commands, memory migration/search/backfill/mirroring, event contracts, dashboard sidebar/workspace view, and their focused tests. No high- or medium-risk file was left uncovered.

## Blocking Findings

### ! [requirements/correctness] `agents.workspace_id` is explicitly omitted — `src/lib/overdeck/infra.ts:70-79`, `src/lib/overdeck/infra.ts:272-276`
AC-1 and FR-4 require both `conversations.workspace_id` and `agents.workspace_id` on fresh and migrated databases. The implementation deliberately adds only the conversation column, and the schema test explicitly documents the agent column as deferred. A fresh or upgraded database therefore fails the stated schema contract, and the new workspace view has to fall back to an issue-id guess instead of the required agent/workspace linkage.

### ! [requirements] The two-door guard allows SQL outside the two door files and misses JOIN reads — `scripts/guard-workspace-doors.sh:15-22`, `scripts/guard-workspace-doors.sh:37-43`
FR-3/AC-2 require the lint gate to reject table SQL everywhere except `resolver.ts`, `writer.ts`, and schema/top-up code. The allowlist excludes every `src/lib/workspaces/*.ts` file, so direct SQL added to `rebuild.ts`, `types.ts`, or any future sibling passes. The patterns also omit `JOIN workspaces|projects|project_targets|pinned_docs`, so a direct join outside the doors passes too. The guard test covers only `FROM workspaces` in an unrelated directory and does not lock the actual invariant.

### ! [requirements/correctness] `pan start`/`pan plan` do not guarantee a workspace row before worktree creation — `src/lib/workspace-manager/create.ts:204-225`
FR-6/AC-4 require the issue workspace row to exist before the worktree is created. This path silently skips the write when the project table has not already been boot-seeded, and it catches writer failures and continues into worktree creation. On a fresh database, a CLI-created worktree can therefore exist without a row until a later dashboard backfill, which is the exact ordering the criterion forbids.

### ! [requirements/correctness] The legacy memory-home migration is not wired into boot — `src/lib/workspaces/rebuild.ts:17-23`, `src/dashboard/server/main.ts:655-659`
WI-6 requires `migrateMemoryHomesToWorkspacesOnce()` to run at boot behind the marker. The implementation calls only project seeding and issue-worktree backfill; its own module header labels boot migration as a follow-up. After upgrade, existing issue-keyed observations and status remain under `{projectId}/{issueId}`, while all new readers use `{projectId}/{workspaceUuid}`, so users lose their existing memory until they know to run the admin rebuild command manually.

### ! [correctness/data preservation] Migration cannot merge a legacy home into an active UUID home — `src/lib/workspaces/rebuild.ts:145-160`
`mergeDirectoryEntries()` renames each top-level entry directly onto the destination. If the UUID home has already received a new `observations/`, `pending/`, `summaries/`, or `rag-runs/` directory before migration, renaming the legacy directory onto the existing directory fails with `ENOTEMPTY`. This is especially likely because boot migration is absent. The migration then aborts before FTS re-pointing, so AC-6/NFR-5's idempotent, non-destructive migration does not hold for a realistic upgraded workspace.

### ! [correctness/security] `pan workspace new --isolated` both fails on the normal branch and constructs a shell-injectable command — `src/cli/commands/workspace-scratch.ts:62-77`
The inferred parent is normally the project's currently checked-out branch. `git worktree add <path> <parentBranch>` tries to check that same branch out in a second worktree and fails with `fatal: '<branch>' is already used by worktree ...`; a local reproduction returned exit 128. The command also interpolates the user-controlled workspace name/path and `--parent-branch` into a shell string, so quotes or shell metacharacters can execute arbitrary commands. AC-5's isolated scratch-workspace path is therefore unusable and unsafe; use an argument-vector spawn and create a distinct scratch branch.

### ! [correctness/requirements] Workspace conversation filtering uses cwd containment instead of `workspace_id` — `src/dashboard/frontend/src/components/workspace/WorkspaceView.tsx:120-128`
The default “This workspace” filter includes every conversation whose cwd is beneath the workspace path. A main workspace rooted at `/repo` therefore includes conversations from `/repo/workspaces/feature-*`, even though those rows belong to separate issue workspaces. FR-15/AC-11 require workspace-scoped conversations using the new linkage; the view must filter by `conversation.workspaceId`, with cwd matching only as an explicit legacy fallback.

### ! [correctness/requirements] Query-expansion logs keep writing to the retired issue-keyed home and drop main/scratch logs — `src/lib/memory/query-expansion.ts:246-250`
`resolveRagRunsFile()` now takes a workspace UUID, but this call passes `identity.issueId` and returns early when the issue is null. After a successful migration, each issue prompt recreates `{projectId}/{issueId}/rag-runs`, splitting one workspace's memory across old and new homes; main/scratch prompts never record expansion decisions. This violates FR-7's workspace-keyed layout and undermines migration idempotency.

### ! [requirements/performance] The new search scopes do not use FTS and `--workspace` does not resolve its owning project — `src/lib/memory/cli.ts:71-90`, `src/lib/memory/cli.ts:190-207`
FR-9 requires `--workspace`, `--project`, and `--global` against the existing FTS. Instead, global search enumerates every workspace directory, reads every daily JSONL file into memory, parses all observations, sorts the full corpus, and only then keeps the requested limit. Runtime and memory therefore grow with the complete history rather than the result set. Separately, `--workspace <id>` still searches the default `overdeck` project unless the caller also supplies `--project`, so a valid workspace in any other project returns no results; workspace name resolution from D-3 is also absent.

### ! [requirements] Event schemas accept `workspaceId`, but event producers still omit it — `packages/contracts/src/events.ts:54-58`
D-10/WI-11 require issue-bearing domain-event payloads to carry the workspace id. This change makes the field optional in contracts, but representative production emitters such as `src/dashboard/server/routes/agents/lifecycle-restart.ts:118-136` still emit only `issueId`; the branch-wide producer search found population only for `operator.intervention`. Consumers therefore receive the same payloads as before, and AC-11 is not met.

### ⊗ [security] Pinned docs can escape the project root and inject arbitrary local files into model context — `src/cli/commands/memory.ts:203-210`, `src/lib/memory/injection.ts:485-509`
An absolute path outside the project is converted to a `../../...` relative path, and a relative traversal is accepted unchanged. Injection later resolves that path against the project root without a containment check and reads the resulting file. `pan memory pin /home/user/.ssh/id_rsa` can therefore persistently place a private key or any other readable local file into prompt-time context sent to the configured model provider. Pin creation must reject paths whose resolved target is outside the project's allowed documentation root.

### ! [requirements/durability] Deleting a workspace removes pin rows but leaves their committed mirrors — `src/lib/workspaces/writer.ts:157-169`
FR-10 requires workspace-scoped pins to be removed with the workspace, and FR-11 makes the state-domain mirror the durable representation. `deleteWorkspace()` deletes the SQLite rows inside a synchronous transaction but never calls `unmirrorPin`, so `memory/pins/` retains stale committed descriptors after destruction. The two stores diverge immediately, and any future recovery from the durable state resurrects pins that the workspace deletion was supposed to remove.

## Non-blocking Findings

### ~ [correctness] Malformed JSON silently becomes a destructive default — `src/dashboard/server/routes/workspace-registry.ts:57-64`, `src/dashboard/server/routes/workspace-registry.ts:137-150`
`readJsonBody` converts invalid JSON to `{}`. The archive route interprets a missing `archived` field as `true`, and the favorite route similarly defaults to `true`, so a malformed mutation body succeeds with a state change instead of returning 400.

### ~ [durability] Pin mirror filenames are collision-prone — `src/lib/memory/state-mirror.ts:83-85`
Replacing `/` with `__` makes `a/b.md` and `a__b.md` map to the same committed filename. Pinning one can overwrite the other's durable descriptor, and unpinning either can delete the mirror for both. Use the pin UUID or a collision-resistant encoding/hash in the mirror filename.

## Clean Dimensions

- No use of `execSync` was added to dashboard-server-reachable code.
- JSONL backfill reads transcript files without modifying, truncating, moving, or deleting them.
- The REST mutation routes apply the existing dashboard mutation-origin/token guard.
- `git diff --check origin/main...HEAD` passed.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-1990/.pan/review/agent-pan-1990-review-ae189541/review.md

## Required action

`pan review request PAN-1990 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1990

# Review — PAN-1990 — cycle 2 — 2026-07-29

## Verdict: CHANGES REQUESTED — the boot migration can permanently skip legacy memory, and the containment/cleanup fixes still leave security and durability failures

## Context
- Branch: `feature/pan-1990`
- HEAD reviewed: `326c48a75815bda5dee42508b62ccf3959b93759`
- Base merge-point: `ed7d6766f3ef2a788daf4d9ead51139fc4ef1a65`
- Prior reviewed HEAD: `ae1895419650db87d0d01d9f3d2c41d6f8febe8c`
- PR: `#3277`
- Review cycle: 2
- Coverage: all four high-risk files, all five medium-risk files, the complete `ae189541..326c48a7` fix delta, and the previously-blocked workspace schema/doors/lifecycle, memory migration/search/pins, event propagation, isolated-worktree, and dashboard filtering paths. No high- or medium-risk file was left uncovered.

## Blocking Findings

### ! [correctness/data migration] Boot migration races the issue-workspace backfill, then writes the global completion marker even when rows were unresolved — `src/dashboard/server/main.ts:526-540`, `src/lib/workspaces/rebuild.ts:274-282`
`backfillIssueWorkspaces()` and `migrateMemoryHomesToWorkspacesOnce()` are launched as independent fire-and-forget promises. The migration depends on the backfill having created issue workspace rows, but it can reach `getWorkspaceForIssue()` first, classify every legacy issue home as unresolvable, and still write `.workspace-keyed`. Every later boot then returns at the marker and never retries, leaving the legacy memory permanently invisible to UUID-keyed readers. Chain/await backfill before migration, and do not mark completion while migratable entries remain unresolved.

### ! [correctness/data preservation] Recursive migration corrupts snapshot JSON files when both homes contain the same filename — `src/lib/workspaces/rebuild.ts:158-182`
The fix appends every colliding file as UTF-8 because it assumes all memory-home files are append-only logs. They are not: `status.json` and `health.json` are single JSON snapshots, and pending entries are standalone `.json` documents. If the UUID home has already produced one of those before migration, appending the legacy file creates invalid JSON such as two concatenated objects. Merge rules must distinguish append-only JSONL/log files from snapshots and per-record JSON files; the current implementation turns a valid upgraded workspace into unreadable memory state.

### ⊗ [security] Lexical containment still follows in-project symlinks to arbitrary files outside the project — `src/lib/memory/pin-path.ts:12-17`, `src/lib/memory/injection.ts:494-509`
`resolveContainedPinPath()` checks only the normalized path string. A project path such as `docs/private` that is a symlink to `~/.ssh/id_rsa`, `/etc/hosts`, or another external file passes as `docs/private`; the injection read then follows the symlink and sends the external file into model context. A local reproduction returned `docs/outside` for an in-project symlink targeting `/etc/hosts`. Validate the real path of both the project root and target (and require the target to exist/read as an allowed regular file) before storing or injecting a pin.

### ! [requirements/architecture] The write door now performs a direct `SELECT` instead of reading pins through the resolver — `src/lib/workspaces/writer.ts:161-170`
FR-3 says every read of `projects`/`workspaces`/`project_targets`/`pinned_docs` goes through `resolver.ts`, while every write goes through `writer.ts`. `deleteWorkspace()` directly selects `doc_path` from `pinned_docs`; the lint guard cannot catch it because `writer.ts` is allowlisted. Fetch the workspace pins through `listPinnedDocs('workspace', id)` before the write transaction so the one-read-door invariant remains structural rather than advisory.

### ! [durability] A mirror push failure after workspace deletion leaves stale pins with no retry path — `src/lib/workspaces/writer.ts:168-178`, `src/cli/commands/workspace-list.ts:218-225`
`deleteWorkspace()` commits deletion of the workspace and all pin rows first, then removes each durable mirror. If any state-door commit/push fails, the function throws after the DB has already forgotten both the workspace and its pin list. `pan workspace destroy` has also already removed the worktree, and a retry finds no workspace row, so it cannot discover or unmirror the stale descriptors. This reproduces the original DB/state divergence under a realistic transport failure; use a durable cleanup intent/outbox or another recoverable ordering that retains enough canonical information to retry until the mirror deletion succeeds.

### ! [requirements] `pan memory search --workspace` still accepts UUIDs only, although D-3 requires `<id|name>` — `src/lib/memory/cli.ts:80-100`, `src/cli/commands/memory.ts:24-35`
The FTS and owning-project fixes are correct for a workspace UUID, but resolution still calls only `getWorkspaceById()`, and the CLI/documentation advertise `<id>`. The approved decision explicitly specifies `--workspace <id|name>`. A workspace name therefore silently falls back to the default `overdeck` project and is also passed as a nonexistent UUID filter, returning no results instead of resolving the workspace or reporting ambiguity.

## Non-blocking Findings

### ~ [correctness] FTS summary hits consume the limit and are then silently discarded — `src/lib/memory/cli.ts:96-120`, `src/lib/memory/cli.ts:123-135`
The shared FTS contains observations and daily summaries, but this CLI API returns only `MemoryObservation`. It requests a bounded mixed-document hit list, then tries to resolve every hit by finding `hit.source` in an observation day file; summary hits (`source = 'summary'`) become null and are dropped without fetching replacement hits. A summary ranked above matching observations can therefore reduce the result count or produce an empty result. Filter `doc_type = 'observation'` in the FTS query for this command.

### ~ [correctness] Failed worktree creation leaves an active workspace row behind — `src/lib/workspace-manager/create.ts:204-238`
The row is now correctly created before the git operation, but later path validation or worktree creation failure returns without archiving/deleting that pre-created row. The workspace registry can therefore advertise an active issue workspace whose path was never created. Preserve the required ordering while compensating the row when creation fails.

### ~ [test isolation/performance] Generic event-store appends synchronously query the process-global database — `src/dashboard/server/event-store.ts:40-47`, `src/dashboard/server/event-store.ts:305-350`
`createEventStore(dbAdapter)` is dependency-injected, but workspace stamping bypasses that adapter and opens/queries the global Overdeck database for every issue-bearing event. Unit stores can therefore depend on unrelated live workspace rows, and the hot append path gains an extra synchronous query. The focused event-store suite passed in the current state, but the result is environment-dependent when a matching issue row exists; inject the workspace resolver or stamp at producer/projection boundaries.

### ~ [input validation] Scratch names are still used directly as path and branch fragments — `src/cli/commands/workspace-scratch.ts:57-84`
Argument-vector execution closes the shell-injection path, but names containing separators, `..`, spaces, or ref-invalid characters can create nested/unexpected target paths or fail late in git. Validate or deterministically slug the name before constructing `scratch-<name>` and `scratch/<name>`.

## Resolved From Cycle 1

- `agents.workspace_id` now exists in fresh/top-up schema and is wired through agent row codecs.
- The workspace-door guard now limits its allowlist to the two doors and detects direct `JOIN` access.
- Issue workspace rows are seeded and created before worktree creation instead of silently skipped.
- Query-expansion logs now use `workspaceId`, including main/scratch turns.
- Issue-bearing event-store and agent-projection paths now stamp `workspaceId` when a row exists.
- Isolated scratch workspaces use a distinct branch and argument-vector git execution; the real-git regression test covers the prior collision.
- Workspace conversation filtering now prefers `conversation.workspaceId` with a legacy cwd fallback.
- Search now queries FTS and resolves the UUID workspace's owning project.
- Malformed workspace mutation JSON now returns 400.
- Pin mirror filenames use a hash, removing the reported transliteration collision.

## Verification Performed

- `git diff --check origin/main...HEAD` passed.
- `npx vitest run --configLoader runner tests/unit/event-store.test.ts` passed: 10/10 tests.
- Pinned-doc symlink reproduction confirmed the containment helper accepts an in-project symlink whose real target is outside the project.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-1990/.pan/review/agent-pan-1990-review-326c48a7/review.md

## Required action

`pan review request PAN-1990 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1990

# Review — PAN-1990 — cycle 3 — 2026-07-29

## Verdict: CHANGES REQUESTED — the cycle-three cleanup ordering can still strand durable pin mirrors, and failed setup can delete the registry row for an existing worktree

## Context
- Branch: `feature/pan-1990`
- PR: `#3277`
- HEAD reviewed: `c19e3b1084c00539feda613993f59a7eacee125a`
- Prior cycle SHA: `326c48a75815bda5dee42508b62ccf3959b93759`
- Cycle diff: 24 files, 614 additions, 57 deletions
- Full PR diff: 125 files
- Coverage: the four high-risk and five medium-risk files were already reviewed in the preceding two cycles; this pass rechecked the complete cycle-three fix delta and its tests across correctness, security, requirements, and performance. No high- or medium-risk file is uncovered. Under the cycle-3 convergence gate, blocking findings below are anchored in code added or reordered by `c19e3b10`.

## Blocking Findings

### ! [durability/correctness] A failed mirror push still becomes unretryable because the retry sees the locally deleted file and skips the push — `src/lib/workspaces/writer.ts:177-185`, `src/lib/memory/state-mirror.ts:57-64`
The cycle-three fix moves `unmirrorPin()` before the SQLite transaction and keeps the rows when it throws, but `removeMemoryStateMirror()` deletes the mirror file and commits locally before discovering that `git push` failed. On retry, `existsSync(target)` is false, so `unmirrorPin()` returns without flushing the local-ahead deletion; `deleteWorkspace()` then removes the workspace and pin rows. The remote `overdeck-state` branch can still retain the stale pin descriptor with no canonical row left to drive another retry. This is the exact transport-failure class the fix claims to close, and state pushes have already shown real non-fast-forward/lock failures during this review. Preserve a durable cleanup intent until the deletion is confirmed on the remote write door, or make the retry flush an existing local-ahead deletion even when the mirror file is already absent.

### ! [correctness/requirements] Compensation deletes the workspace row after failures that occur after a real worktree already exists — `src/lib/workspace-manager/create.ts:240-775`, `src/lib/workspace-manager/create.ts:777-783`
The new wrapper calls `deleteWorkspace(createdWorkspaceRowId)` for every `outcome.success === false`, although many errors are accumulated after `git worktree add` succeeded: port assignment, devcontainer rendering, tunnel/Hume setup, and Docker startup can all add `result.errors`, causing line 773 to return `success=false`. The compensation then removes the row while leaving the worktree and branch on disk. The running process loses the FR-6 workspace registry row until a later boot backfill, and retrying creation encounters the existing non-metadata worktree path. Limit compensation to failures proven to occur before any worktree was created, or retain/archive the row and route partial-worktree cleanup through the canonical workspace lifecycle.

## Convergence Note

Both blockers are introduced by the `c19e3b10` cycle diff. The first could not have been raised in cycle 1 because cycle 3 introduced the unmirror-first ordering specifically to fix cycle 2's DB-first durability defect. The second could not have been raised in cycle 1 because cycle 3 introduced the blanket post-outcome compensation wrapper; before this commit, failure left the pre-created row rather than deleting a row for an already-created worktree.

## Non-blocking Findings

### ~ [security] Realpath containment closes the reported symlink escape, but verification and reading remain separate filesystem operations — `src/lib/memory/pin-path.ts:39-55`, `src/lib/memory/injection.ts:498-505`
A concurrent local process could retarget a symlink after `verifyPinPathContainment()` returns and before `readFile()` opens the lexical path. This requires local filesystem write access during the narrow race and is not a remotely exploitable path by itself, so it does not block this cycle. Opening and reading one verified file descriptor would remove the time-of-check/time-of-use gap entirely.

### ~ [performance] Production event appends still perform a synchronous global workspace lookup for every issue-bearing event — `src/dashboard/server/event-store.ts:27-30`, `src/dashboard/server/event-store.ts:251-252`
Resolver injection fixes unit-store isolation, but the production default retains the synchronous global database query on each append/appendAsync/emitOnly/appendOnce call. This was advisory in cycle 2 and the production behavior is not newly promotable in cycle 3. Producer-side stamping or a cached/injected runtime resolver would remove the extra hot-path lookup.

### ~ [correctness] A failed backfill is swallowed before migration starts — `src/dashboard/server/main.ts:538-544`
The marker guard now prevents unresolved homes from being permanently marked complete, so the prior data-loss blocker is fixed. Still, `.catch(...).then(...)` starts migration after backfill rejects, contrary to the comment that migration must wait for successful backfill. This causes avoidable partial scans and warnings but remains retryable because no marker is written while unresolved homes remain.

## Resolved From Cycle 2

- Boot now sequences migration after backfill completion, and the marker is withheld while any legacy home remains unresolved.
- Snapshot/per-record collisions are no longer concatenated into invalid JSON; only `.jsonl` logs are appended.
- Pin creation and injection now use realpath containment and reject out-of-project symlink targets, missing paths, and directories.
- `deleteWorkspace()` reads pins through `listPinnedDocs()` instead of issuing a direct `SELECT` in the writer.
- `pan memory search --workspace` resolves active workspaces by UUID or unique name and reports cross-project name ambiguity.
- Memory CLI FTS searches now restrict results to observation documents, so summaries no longer consume and disappear from the bounded result set.
- Event-store workspace resolution is injectable for isolated stores.
- Scratch workspace names are validated before becoming branch/path components.

## Verification Performed

- `git diff --check origin/main...HEAD` passed.
- Focused cycle-three verification passed: 10 test files, 89 tests.
- Reviewed the full `326c48a7..c19e3b10` source/test delta and confirmed both blockers are in changed cycle-three paths rather than previously-reviewed unchanged code.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-1990/.pan/review/agent-pan-1990-review-c19e3b10/review.md

## Required action

`pan review request PAN-1990 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1990

# Review — PAN-1990 — cycle 4 — 2026-07-29

## Verdict: CHANGES REQUESTED — the new retry path falsely confirms durability after add/commit failures, and partial polyrepo creation still deletes the registry row for real worktrees

## Context
- Branch: `feature/pan-1990`
- PR: `#3277`
- HEAD reviewed: `cc815c9f3bd3ae61b20e78caf6a82a0a44f2eb77`
- Prior cycle SHA: `c19e3b1084c00539feda613993f59a7eacee125a`
- Cycle diff: 9 files, 337 additions, 44 deletions
- Full PR diff: 127 files
- Coverage: all four high-risk and five medium-risk files were covered in the preceding cycles; this pass reviewed every source and test change in `c19e3b10..cc815c9f` across correctness, security, requirements, and performance. No high- or medium-risk file is uncovered. Under the cycle-4 convergence gate, both blockers are anchored in code added by `cc815c9f`.

## Blocking Findings

### ! [durability/correctness] Retry handles a failed push only; a failed add or commit is misreported as durable — `src/lib/memory/state-mirror.ts:65-73`, `src/lib/pan-dir/auto-commit.ts:650-669`
`commitMirror()` throws for any `flushAutoCommits()` error after the file was removed, including `git add` failure, an index lock, commit-hook/signing failure, or `git commit` failure. In those cases there is no local deletion commit to push: the deletion is merely unstaged or staged. On retry the target is absent, so the new branch calls `pushPendingStateCommits()`; `git push` can return success with no ahead commits, and `removeMemoryStateMirror()` then reports success. `deleteWorkspace()` proceeds to delete the workspace and pin rows while the remote mirror still exists, recreating the exact stale-descriptor divergence this cycle is intended to close. The new test mocks only the committed-but-unpushed case and cannot catch this. The retry must first establish whether the deletion is committed, stage/commit it when necessary, and only then confirm that the commit reached the configured remote.

### ! [correctness/requirements] A partial polyrepo creation leaves real repo worktrees on disk but still deletes the workspace row — `src/lib/workspace-manager/create.ts:217-218`, `src/lib/workspace-manager/create.ts:274-320`, `src/lib/workspace-manager/create.ts:790-796`
`worktreeCreated` becomes true only after the entire polyrepo loop succeeds. If repo A's worktree is created and repo B fails, `result.success` is false and the function returns before setting the flag; the outer compensation then deletes the pre-created workspace row even though repo A's worktree and the workspace directory remain. The same happens when one symlink is created before a later repo/symlink fails. A retry collides with the non-metadata workspace path, and the registry remains blind until boot backfill. Set the preservation state as soon as any durable workspace content/worktree is created, or use a canonical cleanup operation that removes every partial child before deleting the row. The added regression test covers only a monorepo failure after the single worktree succeeds.

## Convergence Note

These findings arise from the cycle-four fixes themselves. The first could not have been raised in cycle 1 because `pushPendingStateCommits()` and the target-absent retry branch were introduced by `cc815c9f`; the prior code silently returned instead. The second is in the newly-added `worktreeCreated` compensation guard: cycle 3 introduced blanket compensation, while cycle 4 narrowed it but omitted the partial-polyrepo state.

## Non-blocking Findings

### ~ [security] The combined read still has a path-replacement race between `realpath()` and `open()` — `src/lib/memory/pin-path.ts:69-94`
Opening and reading from one file handle prevents a second lookup after `open`, but `realTarget` is still a pathname obtained by `realpath()` and reopened later. A concurrent local writer can replace that final path with an out-of-project symlink between those calls, and `open(realTarget, 'r')` follows it. This retains the narrow local time-of-check/time-of-use advisory from cycle 3. An `O_NOFOLLOW` open plus handle-based validation, with equivalent protection for replaceable parent components where supported, would make the claim complete.

### ~ [performance/availability] A missing mirror now performs an unconditional remote push attempt — `src/lib/memory/state-mirror.ts:67-72`, `src/lib/pan-dir/auto-commit.ts:650-669`
Even when the pin was never mirrored, `unpin` and workspace cleanup can incur a network push with the full configured timeout. Multiple missing workspace pin mirrors are processed through `Promise.all`, so retries can issue concurrent pushes against the same checkout outside the auto-commit serializer. This is non-blocking relative to the durability defect above, but the retry should be tied to a concrete pending deletion intent rather than every absent path.

### ~ [correctness] Branch-check failure or branch mismatch is treated as a clean no-op — `src/lib/pan-dir/auto-commit.ts:661-669`, `src/lib/memory/state-mirror.ts:68-72`
`pushPendingStateCommits()` returns `null` when `rev-parse` fails or the checkout is on the wrong branch, and the caller interprets `null` as confirmed success. The dedicated state worktree should normally satisfy the invariant, so this does not independently block the cycle, but a durability retry should fail closed on a configured repository whose branch cannot be verified.

## Resolved From Cycle 3

- A locally committed deletion whose first push failed is now retried instead of being skipped solely because the target file is absent.
- Monorepo and fully-created polyrepo workspaces retain their registry row when later setup stages report errors.
- Boot migration now stops when issue-workspace backfill fails instead of continuing through a swallowed rejection.
- Prompt injection reads through one verified file handle, reducing the prior symlink race window and preserving static realpath containment.

## Verification Performed

- `git diff --check origin/main...HEAD` passed.
- `npm run typecheck` passed, including contracts, hooks, evals, dashboard ratchet, and frontend.
- Focused cycle-four verification passed: 5 test files, 46 tests.
- Reviewed the complete `c19e3b10..cc815c9f` source/test delta and confirmed both blockers are in cycle-four changed paths rather than unchanged code previously accepted or left advisory.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-1990/.pan/review/agent-pan-1990-review-cc815c9f/review.md

## Required action

`pan review request PAN-1990 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-1990

# Review — PAN-1990 — cycle 5 — 2026-07-29

## Verdict: CHANGES REQUESTED — the corrected retry cannot reach its push fallback after a committed deletion because `git add` fails on the now-untracked path

## Context
- Branch: `feature/pan-1990`
- PR: `#3277`
- HEAD reviewed: `32d9670e0783cbb846633e49c8edbf111809dd06`
- Prior cycle SHA: `cc815c9f3bd3ae61b20e78caf6a82a0a44f2eb77`
- Cycle diff: 5 files, 170 additions, 32 deletions
- Full PR diff: 127 files
- Coverage: all four high-risk and five medium-risk files were covered in preceding cycles; this pass reviewed every source and test change in `cc815c9f..32d9670e` across correctness, security, requirements, and performance. No high- or medium-risk file is uncovered. The blocker is anchored in cycle-five code under the convergence gate.

## Blocking Finding

### ! [durability/correctness] A committed-but-unpushed deletion now fails at `git add` and never reaches the push-only retry — `src/lib/memory/state-mirror.ts:88-103`, `src/lib/pan-dir/auto-commit.ts:563-580`
The target-absent branch unconditionally calls `flushMirrorCommit()` before `pushPendingStateCommits()`. That repairs a prior add/commit failure while the deleted file is still tracked by `HEAD`, but it breaks the other required case: after the first attempt successfully committed the deletion and only the push failed, the file is absent from the working tree, index, and current `HEAD`. `git add -- <deleted-path>` then exits 128 with `fatal: pathspec '<path>' did not match any files`; `flushMirrorCommit()` returns `errored`, and line 90 throws before the already-created local commit can be pushed. A local real-git reproduction confirmed this exact behavior. The same error occurs for a mirror that genuinely never existed, so no-op unpins and cleanup of missing descriptors also fail instead of reaching the documented fallback. The tests mock `flushAutoCommits()` as `{ committed: false, reason: 'no diff' }`, which is not what real git returns for a path absent from `HEAD`. Distinguish tracked/staged deletions from already-committed or never-tracked paths before choosing commit retry versus push-only retry, and add a real-git regression test for the committed-deletion case.

## Convergence Note

This blocker is introduced by `32d9670e`: cycle 4 added the push-only retry, while cycle 5 inserted the unconditional add/commit attempt ahead of it. The previous cycle could not flag that ordering because `flushMirrorCommit()` was not called for an absent target then.

## Non-blocking Findings

### ~ [correctness/requirements] Polyrepo compensation now preserves the row even when every child worktree fails — `src/lib/workspace-manager/create.ts:278-290`
Setting `worktreeCreated = true` immediately after creating the empty container correctly preserves rows for partial polyrepo success, but it also preserves an active issue row when the first and every child worktree/symlink fails. The empty directory is retryable because the pre-worktree staging helper accepts an empty directory, so this does not block the cycle, but the registry temporarily advertises an unusable workspace. Tracking whether any child was actually created would preserve partial work while still compensating a total failure.

### ~ [security] `O_NOFOLLOW` protects only the final pathname component — `src/lib/memory/pin-path.ts:83-104`
The changed open now fails closed if the final `realTarget` component is swapped for a symlink, resolving the concrete final-component race from cycle 4. A concurrent local writer can still replace an ancestor directory between `realpath()` and `open()`, as the code comment acknowledges. This remains a narrow local-filesystem advisory rather than a remotely exploitable blocker.

### ~ [performance/availability] Every absent mirror first invokes the full commit flush path — `src/lib/memory/state-mirror.ts:88-100`
Even a never-created descriptor now queues a commit, fetches the branch, and attempts `git add` before deciding whether a push retry is needed. Besides the blocking pathspec behavior, this adds avoidable Git/network work to no-op unpins and workspace cleanup. A concrete deletion-intent check should select the recovery path without running both flows speculatively.

## Resolved From Cycle 4

- Retry correctly re-runs add/commit when the first attempt removed the file but failed before creating a commit.
- Partial polyrepo creation now retains the workspace row once the polyrepo container is materialized.
- `O_NOFOLLOW` closes the final-component symlink replacement race during pinned-document reads.

## Verification Performed

- `git diff --check origin/main...HEAD` passed.
- `npm run typecheck` passed, including contracts, hooks, evals, dashboard ratchet, and frontend.
- Focused cycle-five verification passed: 4 test files, 32 tests.
- Real-git reproduction: after committing a file deletion, `git add -- <deleted-path>` returned exit 128 with `pathspec did not match any files`, contradicting the mocked retry test's `no diff` assumption.
- Reviewed the complete `cc815c9f..32d9670e` source/test delta and confirmed the blocker is in cycle-five changed code.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-1990/.pan/review/agent-pan-1990-review-32d9670e/review.md

## Required action

`pan review request PAN-1990 -m "Fixed review issues"`

Issue #3740, opened 2026-08-15
Red main: regenerate composer command manifest after UAT batch


eltmon:
Main CI run 31888209151 is red on merge commit e5a4e170b5.

Root cause: lint:slash-commands regenerates packages/contracts/src/composer-commands.generated.ts and finds drift. The pan handoff command description now says MAX 10000 characters, while the committed generated manifest still says MAX 500 characters.

Required fix:
- Run the canonical slash-command generator.
- Commit only the generated manifest correction required by current main.
- Verify npm run lint:slash-commands and the complete required gate.
- Add or preserve a mechanical guard so future source-command help changes cannot land with a stale generated manifest.

Evidence: CI lint job 95020476536 failed at 2026-08-15T13:53:31Z with composer-command manifest has drifted from the CLI registry.

Flywheel-Affects-Criterion: 2,3,5

eltmon:
Strike fix is implemented and committed locally as 851f770460, and the targeted mechanical guard passes: npm run lint:slash-commands reports the composer manifest matches the CLI registry. The branch is synced with current main. Full typecheck and full lint pass. The complete npm test gate is deterministically red on three failures unrelated to the generated-manifest-only change: two assertions in tests/unit/lib/cloister/deacon-swarm-doneness.test.ts (unexpected issue-level finalization for PAN-900/PAN-902) and one duplicate-surface assertion in tests/unit/lib/overdeck/no-loss-matrix.test.ts (four restart-gate routes). The same three failures reproduce when those two files run alone. Per strike scope, I am not fixing these orthogonal failures or pushing/signaling strike readiness. Unblock by restoring the current main test gate or explicitly authorizing landing despite these known unrelated failures.

eltmon:
Update after syncing three new main commits: the no-loss-matrix duplicate-route failure is fixed upstream and now passes. Two deterministic, unrelated failures remain in tests/unit/lib/cloister/deacon-swarm-doneness.test.ts: the PAN-900 and PAN-902 cases unexpectedly emit issue-level finalization. They still reproduce when the file runs in isolation. PAN-3740 remains committed and clean on strike/pan-3740, with its targeted manifest guard passing. I still need either green main or explicit authorization to land despite these two known unrelated failures.

eltmon:
[...]

PR #3748, opened 2026-08-16
PAN-3740

eltmon:
**Issue:** #3740


eltmon:
Superseded: the composer command manifest was regenerated on main by db474d677cd and c7ca87bb77e; merge-tree against main is a no-op.

Issue #3787, opened 2026-09-08
Direct subagent composer and live working indicators


eltmon:
Add a bottom composer to the selected subagent transcript in conversation views for both Codex and Claude Code.

Requirements:
- Clearly identify the selected child as the message recipient.
- Preserve child history and keep parent and child messages and runtime state isolated.
- Use a verified harness delivery path to send follow-up input to that child; never silently send ordinary user input to the parent or create a replacement thread.
- Handle running, completed, unavailable, and non-resumable children with accurate delivery feedback.
- Validate that the child belongs to the parent conversation before allowing input.
- Preserve all existing viewing, navigation, transcript, and composer behavior.
- Add regression coverage for recipient isolation, status updates, rejection paths, and UI submission; verify actual browser behavior.
- Update user and developer documentation.
- Show the same live “Working for …” indicator used by main conversations inside active child threads for both harnesses; clear it when the child completes or the session ends.

Implementation checkpoint: verify installed harness mechanisms for direct child input. Codex currently has a single-thread app-server adapter; child input requires thread-scoped routing and notification state. Claude Code has native interactive child input, but Overdeck currently owns a PTY rather than a structured child-input connection; confirm the external delivery mechanism before claiming parity.

Related completed read-only extension: PR https://github.com/eltmon/overdeck/pull/3784. This ticket is operator-directed implementation work.

Operator decision: require direct delivery. Do not show any composer for unsupported sessions; do not relay through the main agent. Implementation supports loaded Codex app-server children. Claude Code PTY and Codex TUI sessions remain read-only because their current Overdeck transports expose no verified direct child-input API. Existing hosts must restart to acquire the new operation.

Implementation: https://github.com/eltmon/overdeck/pull/3788, commit e978437f42756bff76791b0899667436a9d7fe35. Full suite: 14,135 backend + 64 required frontend tests passed. Focused frontend 40 and final host 25 tests passed. Build/typecheck/all lint stages passed. Playwright verified rendering, working timer/completion, recipient targeting with controlled transport, acknowledged clear, isolated drafts, and hidden Claude composer. Isolated Node dashboard boot and route checks passed. Ticket remains open pending merge/deploy and close-out.

eltmon:
Shipped in #3788 (`bd61720a7bb`): direct subagent composer (input accepted only for descendant Codex threads) and live working indicators.

PR #3788, opened 2026-09-08
feat(dashboard): direct subagent composer and live working indicators

eltmon:
Subagent views now show the shared “Working for …” activity indicator and a bottom composer only when direct child delivery is available. A running Codex child no longer loses its indicator because the transcript snapshot reports `streaming: false`.

The composer requires a loaded child in the same persistent Codex app-server process. Both the conversation resolver and host validate parent membership; active turns use `turn/steer` with the expected turn ID, and idle children use `turn/start` with the child ID. Child events cannot replace the primary thread, turn, pane output, or activity. Drafts remain separate per child, and failed or uncertain sends retain their text without automatic retries.

Per the operator's explicit decision, unsupported sessions show no composer. Claude Code PTY sessions, Codex TUI sessions, old hosts, and unloaded children remain read-only. There is no parent relay, new replacement thread, or automatic child resume. Existing conversations must restart their host to acquire the new operation. User and developer docs describe this limitation.

Validation:
- Full `npm test`: 14,135 backend tests and 64 required frontend tests passed; 51 backend tests skipped by existing configuration.
- Focused frontend: 40 tests passed, including composer, activity timer, and existing prompt editor behavior.
- Final host isolation tests: 25 passed.
- Typecheck, all lint stages, build, and route/no-loss audits passed.
- Playwright preview: rendered and inspected the composer and working timer; verified child recipient, clear-on-ack, isolated persistent drafts, completion state, and no Claude composer. Browser transport used controlled responses; no messages were sent into unrelated live agent threads.
- Isolated Node dashboard boot returned healthy; GET and POST child-input routes returned the expected missing-conversation 404.

Tracks #3787. Leave the ticket open until merge and deployment satisfy close-out.

### DIFF
--- a/docs/CONVERSATION-SUBAGENTS.md
+++ b/docs/CONVERSATION-SUBAGENTS.md
@@ -98,3 +98,35 @@ Selection accepts only safe IDs and requires membership in the parent's verified
 descendant tree. It never interprets an ID as a path, follows directory/file
 symlinks during discovery, or searches another Codex home. Missing or unrelated
 IDs cannot fall back to the parent transcript. All discovery and reads are asynchronous.
+
+### Child activity
+
+Selected subagent transcripts show the same animated “Working for …” row as
+main conversations, including the tool activity icon. Codex child activity comes
+from its own task lifecycle in the subagent list: Codex transcript snapshots have
+`streaming: false`, so that field cannot override a running child. Claude Code
+also uses the child stream's activity. Ended or disconnected parent sessions do
+not display stale child activity. The main agent's busy state does not determine
+whether a selected child is working.
+
+### Direct child composer
+
+The bottom composer appears only when the live Codex app-server host confirms
+that the selected child belongs to its primary thread and is loaded with an
+idle or active turn. Active turns receive `turn/steer` with the exact expected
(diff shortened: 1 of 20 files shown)


eltmon:
Parked: conflicts with main and PAN-3917 (the Cut) rewrites the surfaces this touches. Rebase onto post-Cut main before review; not landing before the restructure.

eltmon:
Merged main (post-cut), resolved conflicts: `app-server-host.ts` and `app-server-manager.ts` keep main's owner/descendant/foreign thread classification (#4028/#4034). The branch's own thread filters are gone, and child input (`turn/steer`/`turn/start`) now needs `threadScope === 'descendant'`, so owner threads, native-CLI `/new`/`/fork` threads and unannounced threads never accept input. `docs/CONVERSATION-SUBAGENTS.md` keeps both sides and documents the descendant gate and the attached native CLI. `scripts/file-size-allowlist.txt` keeps main's entries and adds the PAN-3787 line again. The `SubagentTranscript` working-indicator fix and #4054's `subagent.status` both apply. Tests were rewritten for the classification. Typecheck, eslint and the affected server and frontend vitest suites pass. Ready for review.

PR #3817, opened 2026-09-09
fix(infra): preserve JSONL transcripts outside vendor cleanup roots

eltmon:
Claude Code's default 30-day transcript cleanup can leave Overdeck conversations with metadata but no history. Add a host service that sets native and managed Claude retention to 365,000 days and takes permanent Restic snapshots every five minutes outside vendor cleanup roots. Older snapshots survive source deletion and truncation; no automated expiry or pruning is installed.

The safeguard is separate from dashboard deployment. It includes exact settings backups, failure reporting, a lock against overlapping jobs, nested agent JSONL coverage, and recovery instructions. A malformed settings file does not prevent the independent backup. The service has been installed and verified on the development host; this PR packages the script, units, tests, and documentation for review. It does not change existing in-flight session-home work.

Validation: four Python tests using real Restic, including recovery after source append/truncation/removal, malformed settings, settings symlinks, unusual filenames, and full fixture repository integrity. Real host snapshots completed and a Tindra transcript restored byte for byte. Systemd units validated. Only Python/scripts/docs changed; the TypeScript application was not rebuilt.

Limits: this is local host protection with up to five minutes of backup latency. It requires Python 3, Restic >= 0.16, and installation of the user systemd units. External disk-loss protection requires backing up both the repository and its password. Overdeck does not automatically render archived Restic snapshots.

### DIFF
--- a/docs/TRANSCRIPT-PRESERVATION.md
+++ b/docs/TRANSCRIPT-PRESERVATION.md
@@ -1,0 +1,92 @@
+# Permanent JSONL preservation
+
+Claude Code deletes old local transcripts at startup. Its default
+`cleanupPeriodDays` is 30; zero is invalid and older versions interpreted zero
+as disabling transcript persistence. Overdeck's conversation metadata can
+outlive the JSONL, leaving a title and cost totals without readable history.
+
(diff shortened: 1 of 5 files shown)


eltmon:
Merged main; ready for review

Issue #3841, opened 2026-09-16
Dashboard button: open a No-Project conversation that reviews recent cross-issue activity and summarizes pipeline problems (retrospective)


eltmon:
## What

Add a dashboard action (Command Deck header or the No Project overview) that opens a new conversation in **No Project** whose kickoff prompt asks the AI to review all recent pipeline activity across issues and produce a retrospective of the problems that occurred, separated into **pipeline/substrate faults** and **work-agent quality faults**, with evidence pointers and the issues already filed for each.

The model for the output is the summary produced by the shepherding conversation for PAN-3836 on 2026-09-16 (conversation 20260916-7665). That summary took a human-directed session hours of manual digging; this button should produce the same shape of report on demand.

## Why

PAN-3836 needed eight rework cycles. Roughly half the delay came from substrate faults (planning auto-handoff lost to a lock race, a latent red test on main hidden by CI's skip_vitest shortcut, review anchors lagging the reviewed commit, a work session wedged by monitor-mail feedback, `pan tell` hangs) and half from a haiku-tier work agent. None of that was visible as a whole anywhere; each symptom surfaced one at a time in feedback files, deacon logs, and review syntheses. An operator should be able to ask "what went wrong today, across everything?" and get one answer.

## Inputs the review conversation should be pointed at

- Per-issue records on `overdeck-state` (`records/*.json`): review cycle counts, verification attempts, `scopeDrift`, `panDoneRecoveredAt`, UAT results.
- Review syntheses under each workspace's `.pan/review/*/synthesis.md` and feedback files under `.overdeck/feedback/`.
- Deacon log (`~/.overdeck/logs/deacon.log`) action lines: recoveries, re-dispatches, nudges, stuck-flags, orphan cleanups.
- Agent lifecycle logs (`~/.overdeck/agents/*/lifecycle.log`) and spawn logs for failed spawns.
- CI run conclusions per PR and the `overdeck/test` status stamps.
- Issues filed in the window (labels `pipeline`, `blocks-main`, `bug`).

## Output shape

1. Headline: issues in the window, how many needed >2 review cycles, how many needed manual intervention.
2. Substrate faults: one bullet each with what happened, the evidence path, whether an issue exists (link) or needs filing.
3. Work-agent quality faults: grouped by tier/model, with the recurring failure classes.
4. Recommended follow-ups, each with a one-line rationale.

## Constraints

- Read-only. The conversation must not run pipeline mutations.
[...]

PR #3844, opened 2026-09-16
PAN-3841

eltmon:
**Issue:** #3841

## Acceptance Criteria

- [ ] Add roles/retrospective.md kickoff template
- [ ] Add prompt-invariant test for the retrospective template
- [ ] Add conversation-retrospective renderer and handler module
- [ ] Add POST /api/conversations/retrospective route
- [ ] Add RetrospectiveButton component with window menu
- [ ] Mount RetrospectiveButton in the Command Deck header row
- [ ] Document the retrospective button, endpoint, and kickoff template

## Summary by CodeRabbit

- **New Features**
  - Added a Pipeline Retrospective action to the Command Deck.
  - Choose a review window of the last 24 hours or 7 days.
  - Retrospectives start a read-only conversation with project evidence, pipeline history, feedback, recovery details, and clearly reported data limitations.
  - Added secure launch handling with authentication and request validation.
- **Documentation**
  - Documented the retrospective workflow, evidence sources, conversation templates, and read-only behavior.


eltmon:
# Review CHANGES REQUESTED for PAN-3841

# Review — PAN-3841 — 2026-09-16T21:55Z

## Verdict: CHANGES REQUESTED — top blocker: new agent-launch endpoint lacks the established authenticated mutation guard

## Context
- Manifest: `.pan/review/agent-pan-3841-review-497a66ea/context.json`
- Branch: feature/pan-3841
- Workspace: workspaces/feature-pan-3841
- HEAD reviewed: 497a66ea5ca99722e1dbae78e60100560c1ddf97
- Cycle number: 1
- Prior cycle SHA: none
- Self-review path (initial verdict: passed) was retracted on operator-authorized reconciliation with the actual convoy specialist reports and the independent quality review at `/tmp/overdeck-quality-review-20260916.md`. The PAN-1981 "convoy disabled" directive in the role text was stale for this dispatch.

## Convoy Status
| Sub-role | Signal | Output | Findings |
| --- | --- | --- | --- |
| security | ready | `.pan/review/agent-pan-3841-review-497a66ea/security.md` | 0 — but missed the missing `rejectUnsafeDashboardMutationRequest` guard; the established sibling-route pattern is auth-gated, and this new route is the only launch-an-agent POST that ships behind `validateOrigin` alone |
| correctness | ready | `.pan/review/agent-pan-3841-review-497a66ea/correctness.md` | 0 blockers, 1 SHOULD advisory (`conv.name` runtime check) |
| performance | ready | `.pan/review/agent-pan-3841-review-497a66ea/performance.md` | 0 blockers, 1 SHOULD advisory (serial `for…of` await) |
| requirements | ready | `.pan/review/agent-pan-3841-review-497a66ea/requirements.md` | 0 in-PR-scope gaps (pre-existing TalkItThrough changes flagged as out of scope) |
| independent | operator quality review | `/tmp/overdeck-quality-review-20260916.md` | 1 P1 + 2 P2 on PAN-3841 |

## Blocking Findings

### [security] Authenticated mutation guard missing on `/api/conversations/retrospective` — `src/dashboard/server/routes/conversations-retrospective.ts:26` (HEAD 497a66ea5ca)
The route gates only on `validateOrigin(request)` and then proceeds to the conversation write door. A non-browser client that can supply a trusted `Origin` header (or `Referer`) reaches `handleConversationCreate` and spawns a model session without proving identity. This departs from the established mutation contract used by sibling launch routes (`rejectUnsafeDashboardMutationRequest`, which accepts either the server-side internal token OR a browser session cookie plus CSRF header). The companion button does not send authenticated headers in the committed HEAD. The security specialist reported 0 findings because it scoped to "what this PR introduced" rather than to "what the established contract requires"; the operator's independent review correctly framed it as a deviation from the family pattern.
**Fix:** import `rejectUnsafeDashboardMutationRequest` from `./dashboard-auth.js`, gate the route on it (return the helper's 403 response on failure), and have the button call `dashboardMutationJsonHeaders('/api/conversations/retrospective')` so it transmits the CSRF header. Add a test that a trusted Origin without credentials is rejected. The work agent has uncommitted working-tree fixes for both halves (button header + route gate) — these need to land as commits on the feature branch.

### [requirements] Retrospective prompt lists only `synthesis.md`; misses the actual report forms — `roles/retrospective.md:28` (HEAD 497a66ea5ca)
The prompt instructs the agent to read `<repo path>/workspaces/*/.pan/review/*/synthesis.md`. Current self-review runs write `review.md`; the convoy writes `correctness.md`, `security.md`, `performance.md`, and `requirements.md`. Both reviewed issues (PAN-3841 itself, and the other issues referenced in the retrospective) have real reports in the omitted locations. Following the stated evidence list therefore drops exactly the ignored findings and contradictory verdicts the retrospective is supposed to surface.
**Fix:** extend the prompt to include `review.md` and the four specialist `*.md` files under `<repo path>/workspaces/*/.pan/review/*/`, tie them to the run/head, and route canonical record reads through the resolver rather than pointing the agent at raw JSON paths. The work agent's working tree does not yet contain this fix.

### [security/correctness] `isRetrospectiveWindow` accepts inherited Object properties — `src/lib/overdeck/conversation-retrospective.ts:38` (HEAD 497a66ea5ca)
`return typeof value === 'string' && value in RETROSPECTIVE_WINDOWS;` — the `in` operator walks the prototype chain. The operator's reproduction: `isRetrospectiveWindow('constructor')`, `isRetrospectiveWindow('toString')`, `isRetrospectiveWindow('__proto__')` all return `true`, then `new Date(NaN)` throws `Invalid time value` in the renderer. The UI's two fixed choices do not trigger this defect today, but a client POST of `{ window: 'constructor' }` reaches the renderer and crashes the kickoff request. 400 instead of a crash.
**Fix:** use explicit equality against the two known keys (or an `Object.prototype.hasOwnProperty.call(RETROSPECTIVE_WINDOWS, value)` own-property check) and continue to return 400 for unsupported values. The work agent's working tree does not yet contain this fix.

## Non-blocking Findings
- `[correctness] ~` — `RetrospectiveButton` lacks a runtime check on `conv.name` before `window.location.assign` (correctness advisory). Defensive guard would prevent `/conv/undefined` if the server contract ever changes. Not promoted to blocker; documented for the next pass.
- `[performance] ~` — `collectRetrospectiveProjects` does sequential `resolveStateReadHomeAsync` per project. Off the hot path; fine for typical 1–5 project counts. Parallelize if the registry grows.
- `[requirements] ~` — `TalkItThrough.tsx` carries pre-existing changes from PAN-3834 inherited via local merge. Out of scope for PAN-3841; flagged so the merge step knows PAN-3834 needs a separate merge path or rebased companion PR.

## Browser UAT
The operator's note: browser UAT remains unproven because pre-existing workspace session/origin faults blocked the live click-to-running flow. The Sonnet test agent was investigating those environment failures at the time of review. Not a PAN-3841 defect; an environmental constraint on the UAT lane. Flagging here so the verification step knows the criterion is unexercised.

## Cycle note
The work agent already has uncommitted working-tree fixes for the P1 (auth guard + client headers). P2 fixes (prompt review-files coverage + window own-property check) are not in the working tree yet. Once the work agent commits and pushes all three, a fresh review cycle can pass cleanly.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3841/.pan/review/agent-pan-3841-review-497a66ea/review.md

## Required action

`pan review request PAN-3841 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-3841

# Review Synthesis — PAN-3841 — 2026-09-16T22:25Z

## Verdict: CHANGES REQUESTED — top blocker: kickoff template instructs a CLI verb (`pan records show`) that does not exist

## Context
- Manifest: `.pan/review/agent-pan-3841-review-0dedaa83/context.json`
- Branch: feature/pan-3841
- Workspace: workspaces/feature-pan-3841
- HEAD reviewed: 0dedaa83d95514e04dc784ca0d31af1b8e9cda45
- Cycle number: 2
- Prior cycle SHA: 497a66ea5ca99722e1dbae78e60100560c1ddf97 (synthesis at `.pan/review/agent-pan-3841-review-497a66ea/`)
- Cycle diff: 7 files / +181 / -22 across the three work-agent fix commits (`6372fff` auth guard, `a6f02a7` window own-property check + body guard, `0dedaa83` prompt coverage of review/specialist reports + records resolver)
- Operator follow-up at `/tmp/overdeck-quality-review-20260916.md` flagged the prior cycle's blockers; rerun confirmed auth and window fixes, then surfaced this cycle's prompt defect independently.

## Convoy Status
| Sub-role | Signal | Output | Findings |
| --- | --- | --- | --- |
| security | ready | `.pan/review/agent-pan-3841-review-0dedaa83/security.md` | 0 blockers, 3 `?` advisories |
| correctness | ready | `.pan/review/agent-pan-3841-review-0dedaa83/correctness.md` | 1 `!` blocker, 3 `~` advisories, 3 `?` notes |
| performance | ready | `.pan/review/agent-pan-3841-review-0dedaa83/performance.md` | 0 blockers, 4 `?` advisories |
| requirements | ready | `.pan/review/agent-pan-3841-review-0dedaa83/requirements.md` | 0 in-PR blockers; 2 Tier-4 UAT observations unevidenced; merge-scope decision flagged |

## Cycle-2 reconciliation

| Prior blocker (cycle 1) | Status at HEAD 0dedaa83 |
| --- | --- |
| Auth guard missing on launch endpoint | ✅ resolved (`6372fff` adds `rejectUnsafeDashboardMutationRequest`; button sends `dashboardMutationJsonHeaders`; six route tests cover foreign-origin, no-credentials, cookie-without-CSRF, internal-token-only) |
| Prompt listed only `synthesis.md`, missed `review.md` + specialist reports | ⚠️ partial — the prompt now references those files, but the new "canonical read door" line instructs `pan records show` and `pan show`, which the cycle-2 correctness reviewer and the operator's follow-up both flag as broken (see Blocking Findings) |
| `isRetrospectiveWindow` accepted inherited Object properties | ✅ resolved (`a6f02a7` uses `Object.prototype.hasOwnProperty.call`; `isRetrospectiveRequestBody` also rejects `null`/arrays/primitives) |

## Blocking Findings

### [correctness] Kickoff template instructs a CLI verb that does not exist — `roles/retrospective.md:27` (HEAD 0dedaa83)
The "Inputs" section now reads: `pan records show <issue-id> --project <project-key>`, or `pan show <issue-id>`. There is no `records` command in the CLI — `src/cli/index.ts` registers no `records` subcommand, and `src/cli/commands/` has no `records.ts`. `pan show <id>` is the runtime lens but does not expose `feedback`, `recoveryTrips`, `scopeDrift`, or `sessionHistory` — the four fields the report's headline numbers ("how many needed more than two review cycles", "how many needed manual intervention") depend on. `tests/unit/prompts/retrospective-template.test.ts` pins the broken string, so the invariant test green-lights the broken instruction instead of catching it. Operator independently reproduced: `pan records show` fails with "unknown command records"; `pan show` does not emit the listed record fields.
**Runtime impact:** every retrospective conversation started from the new button fails at step 1 of its evidence gathering with an unknown-command error and has no documented fallback for the four record fields the report depends on. The agent either invents the headline numbers or silently omits them.
**Fix:** already written in working tree (correctness reviewer: "the working tree at review time replaces input #1 with the real read door — `readIssueRecord` / `readIssueRecordSync` and `getIssueRecordPath` from `src/lib/pan-dir/record.ts`"). The corresponding `tests/unit/prompts/retrospective-template.test.ts` update is also in working tree. **Commit those two files.** Reviews evaluate committed HEAD — as long as the fix sits uncommitted it will keep coming back as a blocker.

## Non-blocking Findings

### [correctness] `~` — `dashboardMutationJsonHeaders` is passed a REST path where it expects a WebSocket RPC URL — `src/dashboard/frontend/src/components/CommandDeck/RetrospectiveButton.tsx:56`
`await dashboardMutationJsonHeaders('/api/conversations/retrospective')` flows into `new URL(resolveRpcUrl(url))`. A relative path has no base, so the URL constructor throws. Latent on the happy path because `ensureDashboardSession` memoizes the connection created at app boot; becomes live if a 5xx-nulls the session and the retrospective click is the first retry. Fix: drop the argument (`dashboardMutationJsonHeaders()`); the corresponding test assertion has to change in lockstep.

### [correctness] `~` — Handler rejections become Effect defects instead of a structured 500 — `src/dashboard/server/routes/conversations-retrospective.ts:33`
`Effect.promise` converts rejections to defects. The sibling `conversations.ts` route avoids this only because `handleConversationCreate` internally try/catches and returns a `jsonResponse` on every path; `handleRetrospectiveConversationCreate` has no such guard and awaits `loadRetrospectiveTemplate` (`readFile`) plus `collectRetrospectiveProjects` → `listProjectsAsync` (can throw on malformed `projects.yaml`). Fix: wrap in `try/catch` returning `jsonResponse({ error }, { status: 500 })`, or use `Effect.tryPromise`.

### [security] `?` — Project config values interpolated into kickoff prompt without newline stripping — `src/lib/overdeck/conversation-retrospective.ts:73`
A newline in `projects.yaml` `key`/`path`/`github_repo` could append lines that read as instructions. Acceptable in the single-operator trust model (source is operator-authored config); fix is defense-in-depth: strip `\r?\n` from each field in `formatProjectLines`.

### [security] `?` — Retrospective route drops the request-provenance log line the sibling create route emits — `src/dashboard/server/routes/conversations-retrospective.ts:31`
Forensics-only gap: a retrospective launched with a leaked CSRF token leaves no `origin`/`referer`/`user-agent`/`x-forwarded-for`/`x-real-ip` trace. Fix: mirror the sibling's one `console.log` line after the guard.

### [security] `?` — No throttle on an endpoint that spawns a billable model session — `src/dashboard/server/routes/conversations-retrospective.ts:24`
Each authenticated POST renders the template and spawns a tmux conversation runtime. Repeated clicks/requests fan out concurrent model sessions and spend. Optional in-flight guard.

### [correctness] `?` — Malformed JSON silently becomes a default 24h retrospective — `src/dashboard/server/routes/conversations-retrospective.ts:32`
`readJsonBody` swallows `JSON.parse` errors and returns `{}`. Matches sibling behavior; flagged only because `isRetrospectiveRequestBody` was added to reject this and cannot see it.

### [correctness] `?` — Route tests do real filesystem and config I/O — `src/dashboard/server/routes/__tests__/conversations-retrospective.test.ts`
The route test mocks the write door but lets the real handler read `roles/retrospective.md` and the machine's real `projects.yaml`. Injecting `loadTemplate`/`collectProjects` (deps already exist) would make it hermetic.

### [performance] `?` — Success path does a full-document navigation instead of client-side routing — `src/dashboard/frontend/src/components/CommandDeck/RetrospectiveButton.tsx:61`
Matches the established `TalkItThrough.tsx` / `ConversationRow.tsx` / `ConversationActionMenu.tsx` pattern; cost is hidden behind the model session start that follows.

### [performance] `?` — Per-project state-home resolution is serial — `src/lib/overdeck/conversation-retrospective.ts:107`
Off hot path; flagged for record. Parallelize via `Promise.all(projects.map(...))` if registry grows.

### [performance] `?` — Kickoff template is re-read from disk on every launch — `src/lib/overdeck/conversation-retrospective.ts:98`
Intentional trade-off (operator edit takes effect on next click). Do not add a cache.

### [performance] `?` — Kickoff prompt directs a full scan of `logs/deacon.log` — `roles/retrospective.md` (Inputs item 4)
Optional `tail`-bounded read for the deacon log; not worth changing on its own.

## Scope Note

- **Pre-existing PAN-3834 ride-along** — `git diff origin/main...HEAD` includes `TalkItThrough.tsx` (+21/-10), its tests, and a docs paragraph from commit `c7e4ba6d78b`, which exists on local `main` but not on `origin/main`. No PAN-3841 commit edits those files; the prohibition in NonGoal 6 was about scope of work performed, and that is respected. **Operator decision:** either push PAN-3834 on its own before this lands, or accept that it rides along here. The requirements reviewer flagged this in detail and it is not a `⊗` from this PR.
- **Cycle-2 convergence gate** is not yet in effect (cycle 2 of 2). The blocker is inside the cycle diff (the prompt edit landed in `0dedaa83`), not a re-litigation of cycle-1 findings.
- **Browser UAT** (retro-mount.ac2 / ac4) is unevidenced at code level — the constituent behaviors are unit-verified. Operator's screenshot of No-project navigation succeeded; harness spawn failed in the workspace due to pre-existing workspace session/origin faults (carried from cycle 1). This is an environmental UAT constraint, not a PAN-3841 defect. UAT lane needs its own recovery.

## Clean Sub-roles

- **security**: clean (3 `?` advisories only).
- **performance**: clean (4 `?` advisories only).
- **requirements**: clean (0 in-PR blockers; the merge-scope note is the operator's call).

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3841/.pan/review/agent-pan-3841-review-0dedaa83/synthesis.md

## Required action

`pan review request PAN-3841 -m "Fixed review issues"`

eltmon:
# Review CHANGES REQUESTED for PAN-3841

# Review — PAN-3841 (cycle 5) — 2026-09-17T01:30Z — RETRACTED PASS

## Verdict: CHANGES REQUESTED — top blocker: cycle-5 passed verdict was wrong; operator's required corrections from 01:04:52 were not addressed in this cycle's commits

## Context
- Manifest: `.pan/review/agent-pan-3841-review-ffb32f54/context.json`
- Branch: feature/pan-3841
- Workspace: workspaces/feature-pan-3841
- HEAD reviewed: ffb32f54751e36d9b70d79f850df62bb61936161
- Cycle number: 5 — passed at 01:25Z, retracted at 01:30Z on operator-authorized re-evaluation.
- Self-review at 01:25Z was wrong because I only read the surface of the new code (the renderer's per-row caps + the prompt rewrite) and did not validate the four required corrections from operator feedback at 01:04:52, did not read `src/lib/pan-dir/record-list.ts` (the real canonical enumeration), and did not exercise the real resolver through tests (the unit tests inject `listRecords` mocks, so the per-project `try/catch` in the collector never sees the resolver's silent failures).
- Probe at `/tmp/pan3841-evidence-probe.ts` exercises the actual collector + formatter with inert fixtures and demonstrates all five defects from the operator's required-corrections list. I re-read every result below against the committed code at HEAD `ffb32f54`; every defect reproduces.

## Retraction note

My cycle-5 verdict was "PASSED" and I should not have signaled it. The operator's earlier feedback explicitly listed required corrections; cycle-5's commits landed the same `bf27cbe09fc feat(cloister)` rename of the architectural fix plus a docs commit, but did not touch any of the five required corrections. I treated the cycle-5 commits as "no new substantive code" and let the prior `?` advisories stand. That was wrong — those items were required changes, not optional caps, and they are still missing from the committed code.

## Required-correction mapping (operator 01:04:52 feedback vs HEAD `ffb32f54`)

### Required #1: Global serialized-size budget
**Status: MISSING.** The renderer caps per-row (`EVIDENCE_LIMITS`) but does not enforce a global byte cap on the rendered `formatEvidence` output. Probe result: a single record with a 1 MB `verificationNotes` string renders `formatEvidence(result)` to **1,000,939 bytes**. `handleConversationCreate` enforces a 50,000-char cap on the message (`conversation-runtime.ts`), so this output would be truncated or rejected downstream — but the renderer itself never enforces the bound, never drops offending records, and never discloses the truncation.

### Required #2: Bounded resolver I/O with surfaced diagnostics
**Status: MISSING.** `src/lib/pan-dir/record-list.ts:33-37` (the canonical enumeration the bridge calls via `listRecordsThroughReadDoor`) does:
```ts
await Promise.all(entries.filter((entry) => entry.isFile() && entry.name.endsWith('.json'))
  .map(async (entry) => { … JSON.parse … }))
```
Unbounded `Promise.all` over every `.json` file in the records directory. The workspaces branch (`record-list.ts:38-48`) does the same over every per-workspace `.pan/records/` directory — no cap, no batching, no sequencing. Two additional silent swallows on the same file:
- `record-list.ts:20` — `fsp.readdir(recordsDir, { withFileTypes: true }).catch(() => [])` — readdir failure produces empty entries, never surfaces as an error.
- `record-list.ts:27-32` — parse failures are `console.warn(...)`-only; the file is dropped silently.

Net effect on the bridge: `collectRetrospectiveEvidence` wraps `await input.listRecords(project)` in `try/catch` and renders an "EVIDENCE UNAVAILABLE" block on rejection. But the production `listRecordsThroughReadDoor` adapter swallows readdir errors and returns `[]`; it never throws. So the collector sees an empty record list, the per-project block renders "No issue records were updated in this window", and the operator has no signal that the canonical door itself failed. The probe's fixture mocks `listRecords` so the unit tests cannot reproduce this — they never call `listIssueRecords`.

### Required #3: `start <= updated <= now`
**Status: MISSING.** `isRecordInWindow`:
```ts
return parsed >= windowStart.getTime() ? 'in' : 'out';
```
Only the lower bound is checked. Probe result: `updated:'2099-01-01T00:00:00Z'` against `windowStart:2026-09-16T00:00:00Z` returns `'in'` and the future-dated record is included in the rendered snapshot. No upper-bound check against `now`. Clock-skewed or future-dated records silently enter the window.

### Required #4: Deterministic recent-first nested selection
**Status: MISSING.** `projectIssueEvidence` slices without sorting:
```ts
sessionHistory: sessions.slice(0, EVIDENCE_LIMITS.maxSessionsPerIssue).map(...)
```
The canonical append-only history is stored oldest-first (event-0 → event-20). `slice(0, 20)` keeps the **oldest 20** events and discards the newest (`event-20`). Probe result: `sessionHistory[0].reason === 'event-0'` and `newestSessionKept === false` for an `event-20`. The top-level `kept.sort(...)` over records is correct (newest record first), but the nested cap is wrong. Same problem applies to `feedback.slice(0, ...)` if the feedback list is oldest-first; `recoveryTrips.slice(0, ...)` if trip order is oldest-first.

### Required #5: Isolated temporary-filesystem fixtures exercising the REAL resolver for migrated and legacy layouts
**Status: MISSING.** The unit test (`src/lib/overdeck/__tests__/conversation-retrospective.test.ts`) injects `listRecords` via the `deps` parameter — every assertion goes through a mock, never through `listRecordsThroughReadDoor` and never through `listIssueRecords`. The cycle-5 test additions continue this pattern. There is no integration test that:
1. Creates a real temp filesystem with `<project>/.pan/records/*.json` (legacy layout) and `<stateRoot>/records/*.json` (migrated layout).
2. Calls the real resolver via `listRecordsThroughReadDoor`.
3. Exercises the collector's failure path: a missing records dir, a corrupt JSON file, a future-dated `updated`, a 1 MB notes string, a 21-element sessionHistory.

Without those tests, the cycle-5 unit suite is green while the production code path has the four defects above.

## Other defects the probe surfaced

### `scopeDrift.declaredScopeUntouched` truncation is underreported
CodeRabbit flagged this as `?` in cycle 5; the probe confirms it. `truncated: Math.max(0, (drift.outsideDeclaredScope ?? []).length - EVIDENCE_LIMITS.maxScopeDriftFiles)` only counts `outsideDeclaredScope`. A `declaredScopeUntouched` array of 16 entries with a 15-entry cap produces `truncated: 0` despite one entry being dropped. The rendered snapshot discloses the kept count but not the dropped count for `declaredScopeUntouched`. Cycle-5 review treated this as cosmetic; the operator's correction raises it to a required change because the disclosure it carries is wrong.

## What needs to happen

1. **Required #1 — global serialized-size budget**: add a byte cap (likely a few hundred KB) in `formatEvidence`; drop records in priority order (oldest first, already-undated) when the cap is hit; surface the dropped count in the rendered snapshot.
2. **Required #2 — bounded resolver I/O + surfaced diagnostics**: replace the unbounded `Promise.all` in `record-list.ts` with a bounded concurrency (e.g. `pMap(records, fn, { concurrency: 8 })`) and a per-project workspace cap. Surface readdir and parse failures through the return value (a discriminated union `{ ok: true, records } | { ok: false, error }`) rather than swallowing them. The collector's existing `try/catch` then sees a rejected promise and renders "EVIDENCE UNAVAILABLE".
3. **Required #3 — `start <= updated <= now`**: extend `isRecordInWindow` to a `'future' | 'in' | 'out' | 'undated'` verdict and count future-dated records as a separate omission in the rendered snapshot.
4. **Required #4 — deterministic recent-first nested selection**: sort `feedback`, `sessionHistory`, and `recoveryTrips` newest-first before slicing. Document the canonical order contract so the read door and the renderer agree.
5. **Required #5 — real-resolver fixtures**: add a new test file that uses `node:os.tmpdir()` (or `fs.mkdtemp`) to spin up isolated temp trees for both layouts, writes real record JSON files via `node:fs/promises`, calls the real `listRecordsThroughReadDoor`, and asserts the per-project `unavailable` branch, the `outOfWindow`/`undated`/`truncated` counts, and the global byte budget.

## Notes for the operator / merge step

- **My cycle-5 verdict was wrong; the operator's re-evaluation is correct.** Required corrections are still missing. The five required changes need to land in `bf27cbe09fc` (or a follow-up commit) and be tested against the real resolver with temp-filesystem fixtures, not against `listRecords` mocks.
- **PAN-3834 ride-along**: unchanged from earlier cycles.
- **Browser UAT**: still unevidenced; environmental lane constraint.
- **Coordinator action**: I am sending `pan tell conv-20260916-1699` with this reassessment and asking the implementation worker to address the five required corrections. Per the operator's directive, the worker fixes **after** this review is recorded, so the blocker is durable in the journal before any rework commit lands.

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3841/.pan/review/agent-pan-3841-review-ffb32f54/review.md

## Required action

`pan review request PAN-3841 -m "Fixed review issues"`

Issue #3842, opened 2026-09-16
Warn when a tiered-execution tier is configured with a model unsuited to that tier


eltmon:
## What

When `tiered_execution.tiers.<tier>.model` (or a role model that feeds a tier) resolves to a model whose capability class does not match the tier's difficulty band, surface a warning: in `pan doctor`, in the Settings page next to the tier picker, and once at spawn time in the agent's spawn log.

Examples of mismatches to flag:

- A `complex`/`expert` tier pointed at a small or fast model (haiku-class, mini-class).
- A `trivial`/`simple` tier pointed at a frontier model (cost warning, not a blocker).
- A tier model whose harness cannot run the configured role (for example a model routed to a harness that lacks subagents when the tier's role needs them).
- A tier model that is not in the available-models catalog or has no provider credentials.

## Why

On 2026-09-16 PAN-3836 (a 14-item feature with routes, a page, a hook, and CLI) ran on `claude-haiku-4-5` because the first xBRIEF item was rated `simple` and the `trivial-simple` tier mapped to haiku. The agent finished the items but then needed six review cycles plus a UAT rejection, each requiring a human to hand it the root cause. The operator has since retuned the tiers, but nothing in the product would have told them the mapping was risky before an issue paid for it.

## Shape

- A capability-class table per model (frontier / workhorse / small) lives beside the model catalog so the check has one source of truth; do not hardcode model ids in the checker (rule: no-hardcoded-model-fallbacks).
- Warning, never a hard error: the operator may choose a cheap model deliberately.
- `pan doctor` output and the Settings page share one checker function.

## Related

PAN-3836, the tiered execution config in `~/.overdeck/config.yaml`, `src/lib/agents/tier-table.ts`.

eltmon:
🤖 **Agent completed work:**

Tier fitness warnings landed across all three surfaces (pan doctor, Settings crew/supervisor badges, spawn log on both registered-slot and single-work paths). Pure checker + node-side context builder share one capability-class table that lives next to the model catalog. Cost mismatches warn rather than block. Docs cover classes, bands, and the three surfaces.

eltmon:
🤖 **Agent completed work:**

[...]

PR #3851, opened 2026-09-16
PAN-3842

eltmon:
**Issue:** #3842

## Acceptance Criteria

- [ ] Add the model capability-class table module with a catalog completeness test
- [ ] Add the pure checkTierFitness / checkStaffingFitness checker with band constants
- [ ] Add buildTierFitnessContextSync, the node-side context builder over the real catalog
- [ ] Add the pan doctor 'Tiered execution' row that validates the loader and reports fitness warnings
- [ ] Log a [spawn] tier fitness line on the registered-slot spawn path
- [ ] Log a [spawn] tier fitness line on the single-work path against the plan's maximum difficulty
- [ ] Render fitness badges on Settings crew rows and the supervisor block from the unsaved draft
- [ ] Document capability classes, difficulty bands, and the three warning surfaces

## Summary by CodeRabbit

* **New Features**
  * Added model capability classifications to help match models with work difficulty.
  * Settings now displays warnings when assigned models are underpowered, overpowered, unavailable, or unsuitable for supervisor duties.
  * `pan doctor` reports tiered-execution fitness warnings and configuration issues.
  * Spawn-time warnings identify model fitness problems without blocking work from starting.
  * Added support for additional model providers and runtimes.

* **Documentation**
  * Updated tiered-execution, model-routing, and troubleshooting guidance with fitness warning details.


eltmon:
# Review CHANGES REQUESTED for PAN-3842

# Review Synthesis — PAN-3842 — 2026-09-16T21:55Z

## Verdict: CHANGES REQUESTED — two slot-path regressions in `41be7f9`: fitness checks the pre-staffing model, and an explicit `--model` slot spawn can now abort

## Context
- Manifest: `.pan/review/agent-pan-3842-review-d967cb97/context.json`
- Branch: `feature/pan-3842`
- Workspace: `/home/eltmon/Projects/overdeck/workspaces/feature-pan-3842`
- HEAD reviewed: `41be7f9ba97dc575a5b678e29aed38a92c5e93eb`
- Convoy HEAD: `d967cb9716a8bcafff9f0469c71259b344bd4fe4` (the four reviewers reviewed this; `41be7f9` landed after they started and was reviewed by synthesis)
- Cycle number: 1
- Prior cycle SHA: none

## Convoy Status
| Sub-role | Signal | Output | Blocking findings |
| --- | --- | --- | --- |
| security | ready | `security.md` | 0 |
| correctness | ready | `correctness.md` | 0 |
| performance | ready | `performance.md` | 0 |
| requirements | ready | `requirements.md` | 0 |
| *synthesis (41be7f9 delta)* | n/a | this file | 2 |

All four reviewers returned zero blockers against `d967cb97`. Both blockers below are in
`41be7f9`, the operator-directed explicit-override fix, which no convoy reviewer saw.

## Blocking Findings

### [synthesis] Slot-path fitness now checks the parent default instead of the staffed model — `src/lib/agents/spawn.ts:139`
**Evidence tier:** Tier 1 (static, sequential assignment) + Tier 2 (existing test pins the divergent staffed value)

`41be7f9` changed the slot call site from `model: tierParams.model` to `model: slotModel`. But
`slotModel` is only assigned the staffed model on the line *after* the log call:

```ts
let slotModel = selectedModel;                       // = determineModel({ model: options.model })  — parent default
const tierParams = resolveSlotTierSpawnParams(...);
logTierFitnessAtSpawn(..., { model: slotModel, ... }, ...);   // ← logs the PRE-staffing model
if (tierParams.model) {
  slotModel = determineModel({ model: tierParams.model, ... }); // ← staffed model assigned here, too late
}
```

**Failure scenario.** Tiered execution on, slot item difficulty `expert`, no `--model`.
`resolveSlotTierSpawnParams` returns `model: 'claude-opus-4-8'` (pinned by the existing test at
`dispatch-tier.test.ts:171`), so the agent genuinely spawns frontier-class and is fit. The fitness
check instead runs against `selectedModel` — the `roles.work` default, typically workhorse-class —
and emits a spurious `underpowered` warning naming a model that is not being spawned, which no
operator action can clear. The inverse is worse: when the tier table staffs a *small* model on an
expert item while `roles.work` is frontier-class, the real underpowered condition is now silently
missed. This breaks FR-5 and the documented contract that spawn-time checking evaluates the staffed
model, on the ordinary (non-override) path — the primary case the feature exists for.

The explicit-override half of the fix is correct here: with an override, `tierParams.model` is
undefined, `slotModel` stays `options.model`, and the check does hit the final model. Only the
non-override path regressed.

**Not caught by tests.** All four new cases call `logTierFitnessAtSpawn` directly with a hardcoded
model literal (`dispatch-tier.test.ts:561,580,510,535`); none exercises the `spawn.ts` wiring, so the
suite is green (56 passed) with the call site wrong.

**Fix.** Move the `logTierFitnessAtSpawn` call below the `if (tierParams.model)` block so it sees the
final `slotModel`, or pass `tierParams.model ? determineModel({ model: tierParams.model, role, spawnKey: modelSpawnKey }) : slotModel`.
Add one call-site test that drives the ordering and asserts the *staffed* model is the one checked —
the current tests cannot catch this class of defect.

**Unaffected:** the single-work path (`spawn.ts:640`) is correct. `selectedModel` is computed as
`determineModel({ model: singleTierParams.model ?? options.model })`, genuinely final in both cases,
and slot spawns short-circuit via `singleTierParams = {}` → empty `planDifficulties` → logger
early-return, so there is no double-log.

### [synthesis] An explicit `--model` slot spawn can now abort on a plan/slot desync — `src/lib/agents/spawn-prep.ts:274-286`
**Evidence tier:** Tier 1 (static, control-flow reordering) + Tier 2 (existing test pins the throw)

Before `41be7f9`, `if (explicitModel) return {}` was the **first statement** of
`resolveSlotTierSpawnParams` — an explicit override bypassed the plan read, the tier resolution, and
the missing-item check entirely. The fix moved that branch to **after** the item lookup, so the
override path now traverses this:

```ts
const item = doc.plan.items.find((c) => c.id === slotItemId);
if (!item) {
  if (explicitTableActive) {
    throw new Error(`Tiered execution is enabled but item '${slotItemId}' was not found ...`);
  }
  return {};
}
if (explicitModel) { ... }   // ← override branch now sits BELOW the throw
```

**Failure scenario.** Tiered execution enabled, a registered slot whose `slotItemId` is no longer in
the plan (an ordinary slot/plan desync — item renamed, removed, or the plan re-finalized while slots
stay registered), and the operator runs a slot spawn with an explicit `--model`. Previously this
returned `{}` and the spawn proceeded on the explicit model. Now `resolveSlotTierSpawnParams` throws
out of `spawnRunWithoutConsentClaim` and **the spawn aborts**. The throw is pinned by the existing
test at `dispatch-tier.test.ts:207-211`, which only exercises the no-override call and so stays green.

This is the one outcome FR-7 and NonGoal 1 forbid most directly — warning-only work must never abort
a spawn — and it is exactly the "preserve override precedence" property the operator asked me to
verify: model-selection precedence survives, but the override's *error bypass* does not.

**Fix.** Extract the difficulty without letting the throw reach the override path — hoist the
`if (explicitModel)` branch above the `if (!item)` block and read `difficulty` from an optional
lookup (`doc.plan.items.find(...)?.metadata?.difficulty`), so a missing item degrades to "no
difficulty, no warning" rather than an exception. If instead the team decides an override spawn
*should* fail loudly on a desync, that is a deliberate behavior change that needs its own test and a
docs line — it must not arrive as a side effect of a warning feature.

## Operator-Directed Item — verified
The operator's independent quality report (P2, `/tmp/overdeck-quality-review-20260916.md`) is
otherwise **resolved** by `41be7f9`:
- Fitness now assessed against the final selected model on both paths — **yes on single-work, broken on slots** (blocker above).
- Override precedence preserved — **partially.** *Model selection* precedence is intact: both resolvers still leave `model`/`harness` unset under an override, so `options.model` continues to win (`spawn-prep.ts:290-296,374-386`, `dispatch-tier.test.ts:194`). The override's *error bypass* is not — see the second blocker. The single-work resolver is unaffected (no throw in that function; `spawn-prep.ts` throws only at `:283`).
- Warning-only preserved (NonGoal 1 / D8) — **yes.** No new throw or error path; the logger stays `console.warn` inside try/catch.
- Explicit small-model / expert-plan tests present on both paths — **yes** (`dispatch-tier.test.ts:500,545`), plus fitting-model negative cases at `:524,572`. The old "override suppresses the warning" assertions were replaced, not deleted.

## Non-blocking Findings

Recommended for the **same pass** (the work agent is already in these files) — two reproduced
warning-accuracy defects that make the feature give wrong advice:

- **[correctness] `~` `provider-not-enabled` fires permanently for every xAI / Groq / Cerebras / Mistral model** — `src/lib/agents/tier-fitness-context.ts:15`. Tier 2 reproduced: `config.enabledProviders` can never contain those four names on any code path, so the warning appears on all three surfaces with no setting that clears it. Make the check one-sided against representable provider names.
- **[correctness] `~` Deprecated ids in the class table misreport `glm-4.7` / `glm-4.7-flash`** — `src/lib/model-capability-class.ts:11-14,48,69`. Tier 2 reproduced: both resolve to `glm-5.1` and report "frontier-class … would cost less", telling the operator to downgrade a model the table itself classes `small`. Look up the literal id before the deprecation hop.

Lower priority:

- **[correctness/requirements] `~` Duplicate React `key` on same-code warnings** — `CrewRow.tsx:57`. A distribution tier with two small-class models yields two `underpowered` warnings keyed identically. Key on `${code}:${model}`.
- **[requirements] `~` `pan doctor --strict` now exits 1 on ordinary configs** — mechanism pre-existing (`doctor.ts:1006`); no CI or hook gate invokes it, so nothing flips red. Noted so the operator is not surprised.
- **[requirements] `~` `CheckResult` re-declared rather than imported** — `doctor-tier-fitness.ts:1-8`. Deliberate, documented in-file; the PRD's literal import created a module cycle. No requirement affected.
- **[correctness] `?` `difficulties` unguarded at the frontend boundary** — `tier-fitness.ts:105,119`. Not reachable today; `?? []` removes the class.
- **[synthesis] `~` Override branch duplicates the pending/`planDifficulties`/`planItems` aggregation** — `spawn-prep.ts:374-383` repeats the same three passes the non-override path runs at `:391-399`. Worth folding into one helper while the work agent is already in this function.
- **[performance] `≉` Redundant config re-read per spawn** — `spawn-prep.ts:315`. Third `loadYamlConfig()` in the same path; mtime-cached, two extra `statSync`. Sub-millisecond.
- **[performance] `?` `tierFitnessWarnings` unmemoized per Settings render** with an O(providers × models) `providerOf` scan — `TieredExecutionSection.tsx:155`, `tiered-crews.ts:213-222`. Microseconds at current catalog size.

Demoted by the scope gate:

- `TalkItThrough.tsx`, `TalkItThrough.test.tsx`, `talk-it-through.spec.ts`, `command-deck.module.css` — **[demoted: pre-existing]**. `git diff main...HEAD` is empty for all four; they belong to `c7e4ba6d78b` (PAN-3834), the merge base. The manifest's diff base predates it. Not this PR's scope; correctly not reviewed as such.

## Scope Note
Not a proportionality or requirements-scope event — 2 blockers against a 6-commit PR diff is well
inside the 3x guideline. Recorded only because the reviewed tree moved
mid-cycle: the convoy was dispatched against `d967cb97`, and the operator-directed fix `41be7f9`
landed while the reviewers ran. Their four reports are sound for `d967cb97` — the requirements
reviewer's AC-6 "Implemented" specifically blesses the old `{}`-on-override behavior (PRD hazard H2),
which `41be7f9` deliberately supersedes on operator instruction. Synthesis reviewed the `41be7f9`
delta directly rather than inheriting a pass on unreviewed code.

## Gates Run
- `npx vitest run dispatch-tier / tier-fitness / model-capability-class` → 56 passed, 3 files.
- `npm run typecheck` → passes; frontend-types guard 0 known errors, none new.
- Full `npm test` and `npm run lint` remain the test specialist's gate.

## Clean Sub-roles
security (0 findings, 0 advisories), requirements (0 findings; 21/21 in-PR-scope requirements
implemented, all 7 non-goals respected), performance (0 findings).

Source: /home/eltmon/Projects/overdeck/workspaces/feature-pan-3842/.pan/review/agent-pan-3842-review-d967cb97/synthesis.md

## Required action

`pan review request PAN-3842 -m "Fixed review issues"`

PR #3880, opened 2026-09-17
fix(context): keep line numbers outside file link targets

eltmon:
File links with a `:line` suffix can open empty Cursor buffers because the opener treats the suffix as part of the filename. The universal file-reference rule now requires absolute file-only targets and places line/column information outside the link. Examples also use plain labels and cover paths containing spaces.

Validation: `git diff --check` passed. Ran `pan sync` on the source checkout and verified that the corrected rule appears verbatim in the Claude, Pi, and Codex global context artifacts, with the old line-link instruction absent. New sessions receive the updated guidance.


eltmon:
Duplicate of #3883, which landed the same rule edit (cherry-picked from the primary checkout's local main) as f450dc856a0.

